Issue #15, opened 2016-08-06
Add support for different types of list items


luciapayo:
It would be nice to come up with another list element to show case how to deal with different types of items in a list. Any ideas?

peter-tackage:
@luciapayo It looks like this is done. We can right a separate issue for adding an "ads" card into the stream, right?

luciapayo:
@peter-tackage It is not donde yet since the adapter does not have the logic to load different view holders depending on the type

peter-tackage:
ok

PR #31, opened 2016-08-22
Wrap the Sound objects into DisplayableItems

luciapayo:
This is the first step for #15. It also the first step to make a RecyclerAdapter that is generic and not item based so it can be reused.

### DIFF
--- a/app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
+++ b/app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
@@ -18,9 +18,9 @@ package com.futurice.freesound.feature.search;
 
 import com.futurice.freesound.R;
 import com.futurice.freesound.core.BindingBaseFragment;
+import com.futurice.freesound.feature.common.DisplayableItem;
 import com.futurice.freesound.inject.fragment.BaseFragmentModule;
 import com.futurice.freesound.viewmodel.Binder;
-import com.futurice.freesound.viewmodel.DisplayableItem;
 
 import android.os.Bundle;
 import android.support.annotation.NonNull;
--- a/app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
+++ b/app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
@@ -149,7 +149,7 @@ public final class SearchFragment extends BindingBaseFragment<SearchFragmentComp
     }
 
     private void showResults(@NonNull final List<DisplayableItem> sounds) {
-        if (sounds.isEmpty()) {
+        if (get(sounds).isEmpty()) {
             get(noResultsTextView).setVisibility(View.VISIBLE);
             get(resultsRecyclerView).setVisibility(View.GONE);
         } else {
--- a/app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
+++ b/app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
@@ -19,8 +19,8 @@ package com.futurice.freesound.feature.search;
 import com.futurice.freesound.R;
 import com.futurice.freesound.core.BindingBaseFragment;
 import com.futurice.freesound.inject.fragment.BaseFragmentModule;
-import com.futurice.freesound.network.api.model.Sound;
 import com.futurice.freesound.viewmodel.Binder;
+import com.futurice.freesound.viewmodel.DisplayableItem;
 
 import android.os.Bundle;
 import android.support.annotation.NonNull;
@@ -139,7 +139,7 @@ public final class SearchFragment extends BindingBaseFragment<SearchFragmentComp
         return binder;
     }
 
-    private void handleResults(@NonNull final Option<List<Sound>> sounds) {
+    private void handleResults(@NonNull final Option<List<DisplayableItem>> sounds) {
         sounds.matchAction(this::showResults, this::showNothing);
     }
 
@@ -148,7 +148,7 @@ public final class SearchFragment extends BindingBaseFragment<SearchFragmentComp
         get(resultsRecyclerView).setVisibility(View.GONE);
     }
 
-    private void showResults(@NonNull final List<Sound> sounds) {
+    private void showResults(@NonNull final List<DisplayableItem> sounds) {
         if (sounds.isEmpty()) {
             get(noResultsTextView).setVisibility(View.VISIBLE);
             get(resultsRecyclerView).setVisibility(View.GONE);
--- a/app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
+++ b/app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
@@ -1,0 +1,46 @@
+package com.futurice.freesound.viewmodel;
+
+import com.google.auto.value.AutoValue;
+
+import android.support.annotation.NonNull;
+
+/**
+ * Wraps a model. Convenient for complex presentation layers such as
+ * {@link android.support.v7.widget.RecyclerView} where different types of model are handled.
+ */
+@AutoValue
+public abstract class DisplayableItem<T> {
+
+    // Types
+    public static final int SOUND = 0;
+
+    public abstract int type();
+
+    @NonNull
+    public abstract T model();
+
+    @SuppressWarnings("NullableProblems")
+    @AutoValue.Builder
+    public abstract static class Builder<T> {
+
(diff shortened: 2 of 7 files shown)


luciapayo:
@peter-tackage is this ok to be merged?

peter-tackage:
I'm back on reviewing now, I'm still not convinced about the int vs enum proposition. I'm trying a few things in the branch to see what works best.

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Wraps a model. Convenient for complex presentation layers such as
+ * {@link android.support.v7.widget.RecyclerView} where different types of model are handled.
+ */
+@AutoValue
+public abstract class DisplayableItem<T> {
+
+    // Types
+    public static final int SOUND = 0;
```
peter-tackage: I tried experimenting with the [Android support annotations](https://developer.android.com/studio/write/annotations.html) to see if they could be used to enforce the kind of type safety that I think using plain old ints lacks. I wrote the code block below to test if defining two constants with the same values inside an `@IntDef` would cause a compiler error - sadly it doesn't.

```java
    @IntDef({SOUND, AD})
    @Retention(RetentionPolicy.SOURCE)
    public @interface ItemType {}

    public static final int SOUND = 0;
    public static final int AD = 0;

    @ItemType
    public abstract int type();
```
This code compiles, which is not great for type safety.

In this case, the choice of enum or int is a choice between type safety and simplicity.

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
```diff
@@ -0,0 +1,46 @@
...
+        DisplayableItem<T> build();
+    }
+
+    @NonNull
+    public static <T> Builder<T> builder() {
+        return new AutoValue_DisplayableItem.Builder<>();
+    }
+
+    @NonNull
+    public static DisplayableItem create(@NonNull final Object model,
```
peter-tackage: To restrict the model type to the generic type, this should be:

```java
    public static <T, E> DisplayableItem create(@NonNull final T model,
                                                final int type) {
   // impl
    }
```

peter-tackage: Actually, now that I'm experimenting with this, I'm not so sure that enforcing `DisplayableItem<T>` is useful. I will experiment some more.

luciapayo: This was intended. The methods in the adapter are for plain DisplayableItems, like update(DisplayableItem items) instead of update(DisplayableItem<T> items). If I remember correctly it does not compile if you try to do update(DisplayableItem.<T>create(...)).

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
luciapayo: I first did it with enums but I didn't like the solution and decided to go for the simplicity of the int constants instead.

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
```diff
@@ -0,0 +1,46 @@
+package com.futurice.freesound.viewmodel;
```
peter-tackage: This class shouldn't be in this package if it has feature specific types in it - such as `public static final int SOUND = 0;` Otherwise features such as ads etc would touch to many areas. It should either go into `com.futurice.freesound.feature.search` or be generified.

app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
```diff
@@ -150,7 +150,7 @@ private void showNothing() {
         get(resultsRecyclerView).setVisibility(View.GONE);
     }
 
-    private void showResults(@NonNull final List<Sound> sounds) {
+    private void showResults(@NonNull final List<DisplayableItem> sounds) {
```
peter-tackage: The `get` call should occur earlier; currently if `sounds` was `null` it would throw on the first `if` statement.

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
peter-tackage: I would suggest that moving it into the `feature` package is the way to go. I'm struggling to think of a benefit of sharing (via a common package) a wrapped, generic data type; it doesn't do anything or explicitly express any useful information.

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
peter-tackage: Ok, this is my proposal:

```java
@AutoValue
public abstract class DisplayableItem<T> {

    // List types
    enum Type {
        SOUND,
        AD // this is just for demo only
    }

    @NonNull
    public abstract Type type();

    @NonNull
    public abstract T model();

    @NonNull
    public static DisplayableItem create(final Type type,
                                         @NonNull final Object model) {
        return new AutoValue_DisplayableItem<>(type, model);
    }
}
```

peter-tackage: See my updated comment at the end of the `@IntRange` comment block.

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
peter-tackage: Yep you are right. I got there eventually!

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
luciapayo: What if we want to implement a pager or a list for a different feature?

luciapayo: Common package inside feature is ok?

luciapayo:
@peter-tackage I am getting this compile error:
Error:(27, 38) error: package com.ryanharter.auto.value.gson does not exist

Any ideas? (I already cleaned, invalidated cached and restarted :P)

peter-tackage:
I see the same problem on your branch. I remember seeing something similar before - I'll investigate.

peter-tackage:
@luciapayo Oddly, I think the problem is that your `DisplayableItem` definition is not consistent with its `Builder`. 

> Parameter type of setter method should be int to match getter com.futurice.freesound.viewmodel.DisplayableItem.type
        Builder<T> type(@NonNull Type type);

To fix it, change:
` public abstract int type();`

to

`public abstract Type type();`

app/src/test/java/com/futurice/freesound/feature/search/SearchViewModelTest.java
```diff
@@ -18,13 +18,30 @@
 
 import com.futurice.freesound.feature.analytics.Analytics;
 import com.futurice.freesound.feature.common.Navigator;
+import com.futurice.freesound.network.api.model.Sound;
```
peter-tackage: This class will be broken because of the move to RxJava2.

luciapayo:
@peter-tackage It is already `public abstract Type type();` and not `public abstract int type();` but still does not compile for me :(

peter-tackage:
@luciapayo Looking at your PR, the signature of `DisplayableItem` still has `public abstract int type();`

luciapayo:
@peter-tackage have you pulled?

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
peter-tackage: Hmm, we can see how it works out. I'm thinking of modifying the package structure a bit to make it more suitable for modules in any case.

app/src/main/java/com/futurice/freesound/viewmodel/DisplayableItem.java
```diff
@@ -0,0 +1,48 @@
...
+@AutoValue
+public abstract class DisplayableItem<T> {
+
+    // List types
+    public enum Type {
+        SOUND,
+        AD // this is just for demo only
+    }
+
+    public abstract Type type();
```
peter-tackage: Add `@NonNull`

PR #36, opened 2016-09-01
[CHORE] Use approved, not merge

peter-tackage:


peter-tackage:
Do as I say, not as I do; self-merging.

PR #43, opened 2016-09-04
Remove module instantiation

peter-tackage:
There were deprecated warnings for the module instantiation of static modules; this PR cleans up that and some other deprecation warnings.

### DIFF
--- a/app/src/test/java/com/futurice/freesound/test/rx/IgnoringObserver.java
+++ b/app/src/test/java/com/futurice/freesound/test/rx/IgnoringObserver.java
@@ -1,0 +1,31 @@
+package com.futurice.freesound.test.rx;
+
+import android.support.annotation.NonNull;
+
+import rx.Observer;
+
+/**
+ * An {@link Observer} which swallows all events without side effects.
+ */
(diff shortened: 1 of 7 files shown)


peter-tackage:
Ignore the Travis failures, there's no build script defined for Travis. Travis just seems particularly keen to build despite that.

app/src/test/java/com/futurice/freesound/test/rx/IgnoringObserver.java
```diff
@@ -0,0 +1,47 @@
...
+package com.futurice.freesound.test.rx;
+
+import android.support.annotation.NonNull;
+
+import rx.Observer;
+
+/**
+ * An {@link Observer} which swallows all events without side effects.
+ */
+public class IgnoringObserver<T> implements Observer<T> {
```
tomaszpolanski: Final

peter-tackage: It wasn't my intention to make this class open, and I could go either way on the usefulness of being able to extend it. But I'll take the "closed by default" route and make it `final` as you've suggested.

PR #44, opened 2016-09-06
Adding retrofit 2

tomaszpolanski:


tomaszpolanski:
Regarding testing: they are using internal tools for that. 
Do we want to do it that way as well?

peter-tackage:
Rebase this with the Travis changes, please

peter-tackage:
Regarding testing, normally such refactoring wouldn't reduce coverage, but in this case create an issue for the testing of the interceptor; it's a nice bite-sized task for someone to pick up.

PR #49, opened 2016-09-07
Add screenshot

peter-tackage:


peter-tackage:
Another self merge, I'm preparing for internal launch.

PR #53, opened 2016-09-19
Hide public constructor in utilities classes

peter-tackage:


app/src/main/java/com/futurice/freesound/utils/AndroidPreconditions.java
```diff
@@ -49,4 +49,8 @@ private static boolean isMainThread() {
         return Objects.equals(Looper.getMainLooper(), Looper.myLooper());
     }
 
+    private AndroidPreconditions() {
+        throw new AssertionError("No instances allowed");
```
tomaszpolanski: We will have plenty of those ctors, how about creating special error for that (no need for repeating `"No instances allowed"`)?

peter-tackage: Good idea!

Issue #56, opened 2016-09-24
Can't deploy the 'release' flavor


kypeli:
Build fails when trying trying to deploy the 'release' flavor. Apparently the ProGuard files are not up to date.

> Note: android.support.v4.media.IMediaBrowserServiceCallbacksAdapterApi21: can't find dynamically referenced class android.service.media.IMediaBrowserServiceCallbacks
> Note: android.support.v4.media.IMediaBrowserServiceCallbacksAdapterApi21: can't find dynamically referenced class android.content.pm.ParceledListSlice
> Note: android.support.v4.media.IMediaBrowserServiceCallbacksAdapterApi21$Stub: can't find dynamically referenced class android.service.media.IMediaBrowserServiceCallbacks$Stub
> Note: android.support.v4.media.ParceledListSliceAdapterApi21: can't find dynamically referenced class android.content.pm.ParceledListSlice
> Note: android.support.v4.text.ICUCompatApi23: can't find dynamically referenced class libcore.icu.ICU
> Note: android.support.v4.text.ICUCompatIcs: can't find dynamically referenced class libcore.icu.ICU
> Note: android.support.v7.widget.DrawableUtils: can't find dynamically referenced class android.graphics.Insets
> Note: com.facebook.stetho.inspector.runtime.RhinoDetectingRuntimeReplFactory: can't find dynamically referenced class com.facebook.stetho.rhino.JsRuntimeReplFactoryBuilder
> Note: com.google.android.gms.internal.zzanx: can't find dynamically referenced class sun.misc.Unsafe
> Note: com.google.gson.internal.UnsafeAllocator: can't find dynamically referenced class sun.misc.Unsafe
> Note: com.squareup.picasso.Utils: can't find dynamically referenced class com.squareup.okhttp.OkHttpClient
> Note: okhttp3.internal.platform.AndroidPlatform: can't find dynamically referenced class com.android.org.conscrypt.SSLParametersImpl
> Note: okhttp3.internal.platform.AndroidPlatform: can't find dynamically referenced class org.apache.harmony.xnet.provider.jsse.SSLParametersImpl
> Note: okhttp3.internal.platform.Platform: can't find dynamically referenced class sun.security.ssl.SSLContextImpl
> Note: retrofit2.Platform: can't find dynamically referenced class java.util.Optional
> Note: retrofit2.Platform: can't find dynamically referenced class org.robovm.apple.foundation.NSObject
[...]

PR #47, opened 2016-09-06
Replacing RxJava version 1 with 2

tomaszpolanski:

PR #57, opened 2016-09-24
Update ProGuard files so that the 'release' flavour will compile.

kypeli:
I updated the ProGuard files for OkHttp3, Retrofit2. And then I had to add new files for OkIo and Picasso, which now also started to complain. I put them in separate files for clarity.

This PR fixes https://github.com/futurice/freesound-android/issues/56

**Please note:** I don't have the auth token, so I haven't been able to see the app in action with actual network requests. Thus, I would ask whoever reviews this to pull the branch and see that network requests actually do still work. Thank you :)


peter-tackage:
Hey thanks. Odd that Travis didn't catch this, because it should be building the release and debug builds with each PR. I'll send you the API key.

kypeli:
Actually, it did build, but when you tried to deploy the app, it gave me these errors. 

Thanks for the API key!

peter-tackage:
Ok that makes more sense. Yes, it must have been an oversight from #47. Goes to show we need to get some Espresso tests (#2) happening.

app/proguard-rules/okio-rules.pro
```diff
@@ -0,0 +1,5 @@
+# Okio
+-keep class sun.misc.Unsafe { *; }
+-dontwarn java.nio.file.*
+-dontwarn org.codehaus.mojo.animal_sniffer.IgnoreJRERequirement
+-dontwarn okio.**
```
peter-tackage: Please add a newline here

app/proguard-rules/picasso-rules.pro
```diff
@@ -1,7 +1,2 @@
--keep class com.squareup.picasso.** { *; }
--keepclasseswithmembers class * {
-    @com.squareup.picasso.** *;
-}
--keepclassmembers class * {
-    @com.squareup.picasso.** *;
-}
+#Picasso
+-dontwarn com.squareup.okhttp.**
```
peter-tackage: Do you know what has changed in Picasso that the existing rules are not longer correct?

app/proguard-config.gradle
```diff
@@ -18,10 +18,12 @@ ext {
...
             './proguard-rules/android-support-rules.pro',
             './proguard-rules/autovalue-rules.pro',
-            './proguard-rules/okhttp2-rules.pro',
+            './proguard-rules/okhttp3-rules.pro',
             './proguard-rules/picasso-rules.pro',
-            './proguard-rules/retrofit-rules.pro',
+            './proguard-rules/retrofit2-rules.pro',
             './proguard-rules/retrolambda-rules.pro',
-            './proguard-rules/rxjava-rules.pro'
+            './proguard-rules/rxjava-rules.pro',
```
peter-tackage: We're using RxJava2 now, so the `rxjava-rules.pro` file should be renamed. Looking at the contents, it seems the rules are no longer relevant (certainly the packages have changed and the classes no longer exist). Given that RxJava2 will no doubt be quite volatile for the next months, I would be inclined to rename the file to `rxjava2-rules.pro` and  have the contents only with a comment:
```
# Currently no rules necessary
```

app/proguard-rules/picasso-rules.pro
peter-tackage: New line at end

app/proguard-rules/okhttp3-rules.pro
```diff
@@ -0,0 +1,6 @@
+# OkHttp3
```
peter-tackage: Do you have a source for these rules, or did you derive them? If there's an external source, it's good to add a comment to the file to show where they come from. There's often many varied opinions on the required entries in Proguard rules for each library with which each author saying "This worked for me", so having traceability helps to know about the authority of the source and where to go if/when the library is updated.

kypeli: Yeah, I got the rules from https://github.com/krschultz/android-proguard-snippets/tree/master/libraries which lists a lot of ProGuard rules for various libraries. I didn't check to see if those rules could be cut down though.

Will add a comment to the file with the sources.

app/proguard-rules/picasso-rules.pro
kypeli: I suspect the need came from the rule changes required because of OkHttp3. I added the rules one by one after seeing what the next issue was.

kypeli:
Thanks for the comments! I updated the PR with regards to the comments. Please let me know if something is still needed.

peter-tackage:
@kypeli All good, thanks for contributing! Please use the "Squash and merge" option when you merge.

kypeli:
@peter-tackage Thanks! Merged and branch deleted.

Issue #22, opened 2016-08-14
Only use Stetho in release builds


peter-tackage:
Stetho should only be used in release builds. Any change request should await the delivery of #21 to add debug/release variants.

kypeli:
This item depends on https://github.com/futurice/freesound-android/issues/56.

PR #58, opened 2016-09-24
Install the Stetho interceptor only in the debug build. 

kypeli:
Separate the InstrumentationModule between "debug" and "release" builds and install the Stetho interceptor only in the debug build.

This fixes the issue https://github.com/futurice/freesound-android/issues/22

I am not completely happy by this approach that in the end makes use of copy-pasting the class between different flavours. On the other hand, it is a quite clean solution and affects only the leaf provider. But if you have a more elegant solution (with Dagger 2) in mind, please let me know, as I am keen to learn. 

**Please note:** This PR depends on https://github.com/futurice/freesound-android/pull/57 and that's why this PR also contains the commit from the PR, but https://github.com/futurice/freesound-android/pull/57 should go in first.

### DIFF
--- a/app/src/debug/java/com/futurice/freesound/app/module/InstrumentationModule.java
+++ b/app/src/debug/java/com/futurice/freesound/app/module/InstrumentationModule.java
@@ -17,13 +17,11 @@
 package com.futurice.freesound.app.module;
 
 import com.facebook.stetho.okhttp3.StethoInterceptor;
-import com.futurice.freesound.network.api.FreeSoundApiInterceptor;
 
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-import javax.inject.Named;
 import javax.inject.Singleton;
 
 import dagger.Module;
@@ -32,20 +30,17 @@ import okhttp3.Interceptor;
 import okhttp3.logging.HttpLoggingInterceptor;
 import timber.log.Timber;
 
-import static com.futurice.freesound.app.module.ApiModule.API_TOKEN_CONFIG;
-
 @Module
 public final class InstrumentationModule {
 
     @Provides
     @Singleton
     @ApiModule.NetworkInterceptors
-    static List<Interceptor> provideNetworkInterceptors(FreeSoundApiInterceptor apiInterceptor,
-                                                        HttpLoggingInterceptor loggingInterceptor) {
-        List<Interceptor> networkInterceptors = new ArrayList<>(1);
-        networkInterceptors.add(new StethoInterceptor());
-        networkInterceptors.add(apiInterceptor);
+    static List<Interceptor> provideNetworkInterceptors(HttpLoggingInterceptor loggingInterceptor,
+                                                        StethoInterceptor stethoInterceptor) {
+        List<Interceptor> networkInterceptors = new ArrayList<>(2);
         networkInterceptors.add(loggingInterceptor);
+        networkInterceptors.add(stethoInterceptor);
         return networkInterceptors;
     }
 
@@ -58,17 +53,17 @@ public final class InstrumentationModule {
 
     @Provides
     @Singleton
-    static FreeSoundApiInterceptor provideApiInterceptor(@Named(API_TOKEN_CONFIG) String apiToken) {
-        return new FreeSoundApiInterceptor(apiToken);
-    }
-
-    @Provides
-    @Singleton
     static HttpLoggingInterceptor provideLoggingInterceptor() {
         HttpLoggingInterceptor interceptor = new HttpLoggingInterceptor(
                 message -> Timber.tag("OkHttp").d(message));
         // TODO Headers only: Bug in HttpLoggingInterceptor, uses not public api
         interceptor.setLevel(HttpLoggingInterceptor.Level.HEADERS);
         return interceptor;
+    }
+
+    @Provides
+    @Singleton
+    static StethoInterceptor provideStethoInterceptor() {
+        return new StethoInterceptor();
     }
 }
--- a/app/src/release/java/com/futurice/freesound/app/module/InstrumentationModule.java
+++ b/app/src/release/java/com/futurice/freesound/app/module/InstrumentationModule.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2016 Futurice GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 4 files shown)


app/src/release/java/com/futurice/freesound/app/module/InstrumentationModule.java
```diff
@@ -0,0 +1,66 @@
...
+    @Provides
+    @Singleton
+    static FreeSoundApiInterceptor provideApiInterceptor(@Named(API_TOKEN_CONFIG) String apiToken) {
+        return new FreeSoundApiInterceptor(apiToken);
+    }
+
+    @Provides
+    @Singleton
+    static HttpLoggingInterceptor provideLoggingInterceptor() {
+        return null;
```
peter-tackage: I think it's better not to return `null` from Dagger module provide methods because exposing `null` gets messy when injected and there are ways to structure your modules to support optional dependencies per build type.

In this case, I think the `InstrumentationModule` should expose a provider of `List<Interceptor>` which the `ApiModule` uses to define its `OkHttpClient`. That way you can expose an empty list from `release` and a non-empty list from `debug`.

peter-tackage:
@kypeli I've made a branch `feature/instrumentation-ideas` with some suggestions. Essentially, the API key `Interceptor` should be moved to the `ApiModule` and the "instrumentation" `Interceptor`s be exposed simply a `List<Interceptor>` from the `InstrumentationModule`. This allows the `debug` type to supply the Stetho and Logging `Interceptor`s  and the release type to supply empty lists.

peter-tackage:
Marking as blocked because of the Proguard changes. Feel free to work on it, but we won't merge until #57 is merged.

kypeli:
Thanks for the good suggestion! Yes, it makes total sense and when thinking about it (with some more experience with Dagger 2 under the belt) it makes total sense. This is why I love working on this kind of project! 

I have the fix ready to be pushed, but I don't dare messing with this branch yet as it could conflict the https://github.com/futurice/freesound-android/pull/57 so I will push the updated commit once https://github.com/futurice/freesound-android/pull/57 lands in master and I can rebase.

kypeli:
I took your advice and reworked the PR. Hopefully it's better now. At least to me it's more what I would have liked it to be in the first place.

app/src/debug/java/com/futurice/freesound/app/module/InstrumentationModule.java
```diff
@@ -40,12 +40,11 @@
     @Provides
     @Singleton
     @ApiModule.NetworkInterceptors
-    static List<Interceptor> provideNetworkInterceptors(FreeSoundApiInterceptor apiInterceptor,
-                                                        HttpLoggingInterceptor loggingInterceptor) {
+    static List<Interceptor> provideNetworkInterceptors(HttpLoggingInterceptor loggingInterceptor,
+                                                        StethoInterceptor stethoInterceptor) {
         List<Interceptor> networkInterceptors = new ArrayList<>(1);
```
peter-tackage: It's worth initializing the size to 2 rather than 1, because we immediately add 2 items here.

app/src/release/java/com/futurice/freesound/app/module/InstrumentationModule.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.List;
+
+import javax.inject.Singleton;
+
+import dagger.Module;
+import dagger.Provides;
+import okhttp3.Interceptor;
+
+@Module
+public final class InstrumentationModule {
```
peter-tackage: I don't usually advocate testing of Dagger modules, but in this case there's no other hook to verify. If you agree, I think it's worth adding a test in `testRelease` folder which verifies that each of `List<Interceptor>` provided by the `InstrumentationModule` in for the release build type is empty.

We use the AssertJ assertions.

app/src/debug/java/com/futurice/freesound/app/module/InstrumentationModule.java
```diff
@@ -40,12 +40,11 @@
     @Provides
```
peter-tackage: This file has unused imports.

peter-tackage:
@kypeli Conditionally approved, there's a few things to cleanup + a test to add.

app/src/debug/java/com/futurice/freesound/app/module/InstrumentationModule.java
kypeli: Thanks, missed those.

app/src/release/java/com/futurice/freesound/app/module/InstrumentationModule.java
kypeli: I think testing this is a great idea so will add tests. 

However, I am a bit uncertain about the AppInterceptors provider right now. I assumed it's just a placeholder for now and it's supposed to return something in the future, since right now both `release` and `debug builds both return empty lists. Of course I can add a test for this too in `testRelease` but should it be empty in the future as well?

peter-tackage: The both lists of interceptors returned from the release version of the `InstrumentationModule` should always be empty (hence the test). Yes, for now the app interceptors are also empty in the debug version; it's just an expansion point for instrumentation. We don't really care what (or how many) instrumentation interceptors are used by the debug version, but we want to make sure that the release version doesn't any.

kypeli:
Updated this PR with minor fixes and added the tests as suggested. The `InstrumentationModuleTest` in `testRelease` must be updated if `provideAppInterceptors()` is no longer returning an empty list.

app/src/testRelease/java/com/futurice/freesound/app/module/InstrumentationModuleTest.java
```diff
@@ -0,0 +1,19 @@
+package com.futurice.freesound.app.module;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class InstrumentationModuleTest {
```
peter-tackage: Missing copyright header.

kypeli: Thanks, fixed.

peter-tackage:
All good for me, once the CI has passed.

peter-tackage:
Longest build time ever. It's waiting for the emulator to start, but doesn't actually need the emulator. I should disable that.

kypeli:
Finally. Had to restart the build once as it was unable to start the emulator the first time.

PR #59, opened 2016-09-25
[CHORE] Cleaning up Rx code

tomaszpolanski:


app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
```diff
@@ -75,55 +84,62 @@ public void unbind() {
...
+                    }
+
+                    @Override
+                    public boolean onQueryTextChange(final String newText) {
+                        e.onNext(get(newText));
+                        return true;
+                    }
+                });
+            }
+        }, BackpressureMode.LATEST)
```
peter-tackage: How fast can you type? :-)

app/src/main/java/com/futurice/freesound/feature/home/HomeViewModel.java
```diff
@@ -28,30 +28,28 @@
...
     }
 
     @Override
     public void bind(@NonNull final CompositeDisposable disposables) {
-        disposables.add(openSearch.subscribeOn(mainThread())
-                                  .observeOn(computation())
-                                  .subscribe(__ -> navigator.openSearch(),
-                                             e -> Timber.e(e, "Error clearing search")));
+        disposables.add(openSearchSubject.observeOn(mainThread())
+                                         .subscribe(__ -> navigator.openSearch(),
```
peter-tackage: If we are going to change the thread to main here, Should we add an `AndroidPreconditions.mainThread()` style check to the Navigator methods? Evidently it didn't matter to the `Navigator` which thread it was invoked from before, although possibly there was a race condition if attempting to navigate when the Activity was being destroyed.

app/src/main/java/com/futurice/freesound/viewmodel/Binder.java
```diff
@@ -16,11 +16,13 @@
...
 package com.futurice.freesound.viewmodel;
 
+import android.support.annotation.NonNull;
+
 import io.reactivex.disposables.CompositeDisposable;
 
 public interface Binder {
 
-    void bind(final CompositeDisposable disposables);
+    void bind(@NonNull final CompositeDisposable disposables);
```
peter-tackage: Unlike the `@NonNull`, the `final` won't be enforced from the interface could you remove it.

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
tomaszpolanski: Now we support anyone writing 500 letters/s :D

app/src/main/java/com/futurice/freesound/feature/home/HomeViewModel.java
tomaszpolanski: Can do with Precondition, don't trust Andriod navigation

app/src/main/java/com/futurice/freesound/viewmodel/Binder.java
tomaszpolanski: ok

app/src/main/java/com/futurice/freesound/feature/home/HomeViewModel.java
peter-tackage: Oops, just gave a 👍 to my own comment.

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
```diff
@@ -75,55 +84,62 @@ public void unbind() {
         }
     };
 
-    private void setClearSearchVisible(@NonNull final Boolean clearVisible) {
-        Preconditions.checkNotNull(closeButton, "Close Button has not been bound");
+    private void setClearSearchVisible(final boolean clearVisible) {
+        checkNotNull(closeButton, "Close Button has not been bound");
 
         get(closeButton).setVisibility(clearVisible ? View.VISIBLE : View.GONE);
```
peter-tackage: I don't think we need both the `checkNotNull` and the `get` call on `closeButton`.

peter-tackage: I would keep the `checkNotNull` here because it's more informative. Perhaps we could consider a similar utility `checkIsBound(View v)` which might give us more context to such `NPE`s with the burden of the long worded text. Then again, we might just be better off using ButterKnife.

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
```diff
@@ -75,55 +84,62 @@ public void unbind() {
...
+
         Intent intent = new Intent(context, SearchActivity.class);
         context.startActivity(intent);
     }
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        setContentView(R.layout.activity_search);
+        setContentView(layout.activity_search);
```
peter-tackage: Not sure sure we should be importing`R.id` and `R.layout` like this, because there can be multiple `R` packages. Any reason you've started doing this?

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
```diff
@@ -75,55 +84,62 @@ public void unbind() {
...
         Option.ofObj(savedInstanceState)
               .ifNone(this::addSearchFragment);
 
-        Toolbar toolbar = (Toolbar) findViewById(R.id.toolbar_search);
+        Toolbar toolbar = (Toolbar) findViewById(id.toolbar_search);
         setSupportActionBar(toolbar);
         getSupportActionBar().setDisplayHomeAsUpEnabled(true);
 
-        searchView = (SearchView) findViewById(R.id.search_view);
+        searchView = get((SearchView) findViewById(id.search_view));
```
peter-tackage: We should definitely add a utility for this, it looks pretty hideous.

peter-tackage: I'm getting tempted by ButterKnife here.

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
```diff
@@ -75,55 +84,62 @@ public void unbind() {
...
             return true;
         });
 
     }
 
-    private boolean search(final String query) {
-        searchViewModel.search(query);
-        return true;
+    @NonNull
+    private static Flowable<String> getTextChange(@NonNull final SearchView view) {
```
peter-tackage: Not that familiar with their use cases yet: why `Flowable` rather than `Observable` here? We don't actually expect backpressure from user input, do we?

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
```diff
@@ -75,55 +84,62 @@ public void unbind() {
...
+        return Flowable.create(new FlowableOnSubscribe<String>() {
```
peter-tackage: This is much nicer and reactive than my side-effect version. Do you think there's value in making this a utility, our own RxAndroid2 starter-kit? We could well have a few iterations on these ideas while RxJava2 is new, it could be worthwhile to avoid ending up with old patterns in different parts of the code.

app/src/main/java/com/futurice/freesound/feature/search/SearchViewModel.java
```diff
@@ -96,28 +89,20 @@ void openSoundDetails(@NonNull final Sound sound) {
 
     @Override
     public void bind(@NonNull final CompositeDisposable disposables) {
-        disposables.add(searchTermRelay.subscribeOn(Schedulers.computation())
-                                       .observeOn(Schedulers.computation())
-                                       .map(String::trim)
-                                       .debounce(SEARCH_DEBOUNCE_TIME_SECONDS, TimeUnit.SECONDS)
-                                       .switchMap(getSearchOrClear())
-                                       .observeOn(mainThread())
```
peter-tackage: Hmm, switching threads to do nothing. Lucky RxJava2 has those performance improvements ;-)

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
tomaszpolanski: Because of AS recommendations, the majority of R are from our resources, if we would be using from somewhere else, then we would use it the full package name from those resources

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
tomaszpolanski: Not a bad idea ;)

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
tomaszpolanski: All things that we cannot affect how often the values can be produces, should be made into Flowable.

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
tomaszpolanski: We can talk about how this could be abstracted, have some experience on how it was done in .NET.

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
peter-tackage: OK, I don't see those recommendation yet. But anyway, we can try this.

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
peter-tackage: @tomaszpolanski Change to `Observable` as discussed.

PR #62, opened 2016-09-30
[CHORE] Introducing naming convention

tomaszpolanski:


app/src/main/java/com/futurice/freesound/feature/search/SearchViewModel.java
```diff
@@ -100,16 +100,16 @@ void openSoundDetails(@NonNull final Sound sound) {
...
+                                               .switchMap(this::searchOrClearOnce)
+                                               .subscribe(nothing1(),
                                                     e -> e(e, "Error when setting search term")));
     }
 
     @NonNull
-    private Observable<Unit> searchOrClear(@NonNull final String searchQuery) {
+    private Observable<Unit> searchOrClearOnce(@NonNull final String searchQuery) {
         return (TextUtils.isNullOrEmpty(searchQuery)
                 ? searchDataModel.clear()
```
peter-tackage: Does the naming convention not apply to `Single` at all? I would be glad if that was the case, `clearOnce` etc looks messy.

peter-tackage: If so, it would be worth adding that to the README.md

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -38,7 +38,7 @@
     private final FreeSoundSearchService freeSoundSearchService;
 
     @NonNull
-    private final BehaviorSubject<Option<List<Sound>>> lastResults = BehaviorSubject.create();
+    private final BehaviorSubject<Option<List<Sound>>> lastResultsStream = BehaviorSubject.create();
```
tomaszpolanski: OnceAndStream guarantees that there will be value emitted after subscription. So if BS does not have a default value, it cannot be OnceAndStream.
Additionally, when I've changed it to have a default value, multiple tests were failing as this is not the desired behavior

app/src/main/java/com/futurice/freesound/feature/search/SearchViewModel.java
tomaszpolanski: It shouldn't apply, but here we have `Single` that we change to an `Observable` so the function that returns that converted `Single` should have a suffix

tomaszpolanski:
I've made the changes, additionally I found some cases where usage of `Completable` is better

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
peter-tackage: OK. There's probably a case for it having a default value and changing the tests, but this PR is not the place for it anyway.

app/src/main/java/com/futurice/freesound/feature/search/SearchDataModel.java
```diff
@@ -22,6 +22,7 @@
 
 import java.util.List;
 
+import io.reactivex.Completable;
```
peter-tackage: This case has a few unused imports.

peter-tackage:
@tomaszpolanski , just the unused imports. Also add mention to the README saying that the naming convention doesn't apply to the other Observable variants because their behavior is indicate via their type (e.g. `Single` is once).

PR #63, opened 2016-10-01
[CHORE] Adding Butter Knife

tomaszpolanski:


app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
```diff
@@ -92,18 +96,18 @@ public void onCreate(Bundle savedInstanceState) {
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container,
                              Bundle savedInstanceState) {
-        return inflater.inflate(R.layout.fragment_search, container, false);
+        View view = inflater.inflate(R.layout.fragment_search, container, false);
```
peter-tackage: I know the ButterKnife docs say to do it here, but it feels odd to be doing the binding in a place where the view is being created: `onCreateView` rather than in the callback indicating the View is ready: `onViewCreated`. If there's no hard requirement to do it in `onCreateView`, I would prefer if the binding was deferred until `onViewCreated` for the sake of mixing together the two responsibilities.

peter-tackage: In the document they also use a `Binding.reset()` call in `onDestroyView()`. See http://jakewharton.github.io/butterknife/

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewHolder.java
```diff
@@ -80,12 +93,11 @@ public void unbind() {
     SoundItemViewHolder(@NonNull final View view,
                         @NonNull final Picasso picasso) {
         super(get(view));
-        this.rootView = view;
+        ButterKnife.bind(this, view);
+        rootView = view;
         this.picasso = get(picasso);
```
peter-tackage: I feel like we should decide about how to assign fields in constructors: I've typically written, `this.rootView = view` regardless of whether the field was named the same as the field. I think this makes for more consistent reading. Or at least we should change the name of the parameter so that this convention is more uniformly applied.

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewHolder.java
```diff
@@ -80,12 +93,11 @@ public void unbind() {
...
-        this.titleTextView = get((TextView) view.findViewById(R.id.textView_title));
-        this.descriptionTextView = get((TextView) view.findViewById(R.id.textView_description));
-        this.waveformViewTarget = new WaveformViewTarget(
-                get((WaveformView) view.findViewById(R.id.waveformView_soundItem)),
+        waveformViewTarget = new WaveformViewTarget(
+                ButterKnife.findById(view, R.id.waveformView_soundItem),
```
peter-tackage: Static import

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
```diff
@@ -99,17 +105,16 @@ public static void open(@NonNull final Context context) {
...
-        Toolbar toolbar = (Toolbar) findViewById(id.toolbar_search);
+        Toolbar toolbar = ButterKnife.findById(this, id.toolbar_search);
         setSupportActionBar(toolbar);
         getSupportActionBar().setDisplayHomeAsUpEnabled(true);
 
-        searchView = get((SearchView) findViewById(id.search_view));
-        searchView.setIconified(false);
-        closeButton = (ImageView) searchView
-                .findViewById(android.support.v7.appcompat.R.id.search_close_btn);
+        get(searchView).setIconified(false);
```
peter-tackage: Is the get necessary? Won't the binding fail with an Exception if it doesn't exist? Sure the field is technically nullable because we don't assign as final constructor, but it would take some asynchronous gymnastics for it to ever occur. I think that we are better off not using that check for ButterKnife bound views unless there's an explicit way path for them to be null.

peter-tackage: I wonder if we could start applying the same logic to Dagger injections too, as I understand it they can't be injected as null either - isn't there now some generated code that prohibits null unless otherwise specified?

tomaszpolanski: On thing, if we would forget to add butterknife compiler, then there would no exception in the bind but the thing would be null.
Second thing, if I don't do this, I will have a warning about it being null.
I am opened for suggestions

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
```diff
@@ -99,17 +105,16 @@ public static void open(@NonNull final Context context) {
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(layout.activity_search);
+        ButterKnife.bind(this);
         Option.ofObj(savedInstanceState)
               .ifNone(this::addSearchFragment);
 
-        Toolbar toolbar = (Toolbar) findViewById(id.toolbar_search);
+        Toolbar toolbar = ButterKnife.findById(this, id.toolbar_search);
```
peter-tackage: static import

app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
peter-tackage: Let's leave it for now.

Issue #33, opened 2016-08-30
Move common dependency versions into properties


peter-tackage:
Common dependency versions, such as those used by the Android support library dependencies, should be moved into the `project.ext` for reuse.

Includes:
* Android support
* Dagger 2
* AutoValue
* AutoValue Parcelable
* Firebase
* Stetho

peter-tackage:
This is relevant: https://segunfamisa.com/posts/android-gradle-extra-properties

PR #64, opened 2016-10-02
Extract library versions into ext properties

peter-tackage:
Implementation of #33. Currently I've only done it for the library dependency versions. Should I do the rest too? I've also extracted the versions when they aren't used multiple times, it just seemed neater, although it does add some cognitive load for looking up the version & extra work when adding new libraries (you need to add the version into ext properties list, rather than just copy the suggested gradle dependency).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -80,98 +80,97 @@ task checkstyle(type: Checkstyle) {
 }
 
 ext {
-    // Shared dependency versions
-    androidSupportLibraryVersion = '23.3.0'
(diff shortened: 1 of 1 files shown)


peter-tackage:
I now wondering if the dictionary approach might be better. Like this - https://github.com/AdamLuisSean/AndroidCleanVIPER/blob/master/dependencies/dependencies.gradle

tomaszpolanski:
I don't think it matters, both approached are almost identical

Issue #46, opened 2016-09-06
Add tests for Interceptors


tomaszpolanski:
Retrofit 2 has a better way to use Interceptors but it is now harder to test them.
There is a way how okhttp [tests](https://github.com/square/okhttp/blob/master/okhttp-tests/src/test/java/okhttp3/InterceptorTest.java) it.

PR #66, opened 2016-10-10
Interceptor tests

peter-tackage:
I've written two different ways to test the `FreeSoundApiInterceptor` (see #46 ). I'm hoping to get some feedback on which approach would be preferred. The style can obviously be cleaned up, so please don't pay too much attention to that.

1. Uses a MockWebserver instance and verifies the request sent at a low level (as a part of the query path) - it's more of an integration test than a unit test. 
1. Is a conventional unit test of `FreeSoundApiInterceptor` using mocks and argument captors.

Thoughts?

### DIFF
--- a/app/src/test/java/com/futurice/freesound/network/api/FreeSoundApiInterceptorTest.java
+++ b/app/src/test/java/com/futurice/freesound/network/api/FreeSoundApiInterceptorTest.java
@@ -19,16 +19,24 @@ package com.futurice.freesound.network.api;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.mockito.ArgumentCaptor;
 
 import java.io.IOException;
 
+import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
+import okhttp3.Protocol;
 import okhttp3.Request;
+import okhttp3.Response;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.RecordedRequest;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 public class FreeSoundApiInterceptorTest {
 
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/futurice/freesound/network/api/FreeSoundApiInterceptorTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class FreeSoundApiInterceptorTest {
+
+    private final static String TEST_TOKEN = "test-token";
```
tomaszpolanski: Bad order, it should be `private static final`

app/src/test/java/com/futurice/freesound/network/api/FreeSoundApiInterceptorTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private FreeSoundApiInterceptor interceptor;
+
+    @Before
+    public void setUp() throws IOException {
+        interceptor = new FreeSoundApiInterceptor(TEST_TOKEN);
+    }
+
+    @Test
+    public void interceptor_addsApiTokenQueryParameter() throws InterruptedException, IOException {
```
tomaszpolanski: Replace both exceptions with `throws Exception`

app/src/test/java/com/futurice/freesound/network/api/FreeSoundApiInterceptorTest.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        ArrangeBuilder withEnqueuedMockResponse() {
+            mockWebServer.enqueue(new MockResponse());
+            return this;
+        }
+
+    }
+
+    private class ArrangeBuilder2 {
```
tomaszpolanski: Why 2 arrange builders?

tomaszpolanski:
I think we should go for `MockWebserver` tests.
We don't write so many interceptors and the logic in them is not so complex. 
So testing bigger part of the integration of the interceptors could be more beneficial.

PR #67, opened 2016-10-10
Add LeakCanary

### LINKED ISSUE #20: Add Leak Canary
Evidently, having Leak Canary in the app would be useful to find memory leaks. This should only be in the debug build type.

peter-tackage:
Resolves #20


app/src/main/java/com/futurice/freesound/app/FreesoundApplication.java
```diff
@@ -72,4 +74,11 @@ private void initStetho() {
                       .enableWebKitInspector(Stetho.defaultInspectorModulesProvider(this))
                       .build());
     }
+
+    private void initLeakCanary() {
+        if (LeakCanary.isInAnalyzerProcess(this)) {
```
tomaszpolanski: Why not just invert this condition and put install in the `if`?

peter-tackage: This was actually just copied from the LeakCanary sample code, but good point it does look a little odd.. I'll change it.

Issue #69, opened 2016-10-12
NPE exception when press on search icon


VasileUngureanu:
`
      Caused by: java.lang.NullPointerException: Cannot return null from a non-@Nullable @Provides method
                                                                                at dagger.internal.Preconditions.checkNotNull(Preconditions.java:47)
                                                                                at com.futurice.freesound.app.module.ConfigModule_ProvideApiModuleApiTokenConfigFactory.get(ConfigModule_ProvideApiModuleApiTokenConfigFactory.java:16)
                                                                                at com.futurice.freesound.app.module.ConfigModule_ProvideApiModuleApiTokenConfigFactory.get(ConfigModule_ProvideApiModuleApiTokenConfigFactory.java:7)
                                                                                at dagger.internal.DoubleCheck.get(DoubleCheck.java:46)
                                                                                at com.futurice.freesound.app.module.ApiModule_ProvideApiInterceptorFactory.get(ApiModule_ProvideApiInterceptorFactory.java:25)
                                                                                at com.futurice.freesound.app.module.ApiModule_ProvideApiInterceptorFactory.get(ApiModule_ProvideApiInterceptorFactory.java:9)
                                                                                at dagger.internal.DoubleCheck.get(DoubleCheck.java:46)
                                                                                at com.futurice.freesound.app.module.ApiModule_ProvideApiOkHttpClientFactory.get(ApiModule_ProvideApiOkHttpClientFactory.java:41)
                                                                                at com.futurice.freesound.app.module.ApiModule_ProvideApiOkHttpClientFactory.get(ApiModule_ProvideApiOkHttpClientFactory.java:12)
                                                                                at dagger.internal.DoubleCheck.get(DoubleCheck.java:46)
                                                                                at com.futurice.freesound.app.module.ApiModule_ProvideFreeSoundApiFactory.get(ApiModule_ProvideFreeSoundApiFactory.java:37)
                                                                                at com.futurice.freesound.app.module.ApiModule_ProvideFreeSoundApiFactory.get(ApiModule_ProvideFreeSoundApiFactory.java:11)
                                                                                at dagger.internal.DoubleCheck.get(DoubleCheck.java:46)
                                                                                at com.futurice.freesound.app.DaggerFreesoundApplicationComponent.getApi(DaggerFreesoundApplicationComponent.java:177)
                                                                                at com.futurice.freesound.feature.search.DaggerSearchActivityComponent$1.get(DaggerSearchActivityComponent.java:75)
[...]

PR #71, opened 2016-10-15
Make API key checking stricter and add encrypted secret key to CI builds

peter-tackage:
IMPORTANT: This also changes the format of the `freesound-api.properties` file. People with an existing file will need to remove the quotation marks around the entry in the file.

Attempts to improve the situation referenced in #69


.travis.yml
```diff
@@ -1,35 +1,29 @@
 language: android
```
peter-tackage: This file is auto-generated by Travis tooling when generating the encrypted key, so hence the "unnecessary" formatting diffs.

app/app-config.gradle
```diff
@@ -15,23 +15,42 @@
...
+    String freesoundApiKey = isCiBuild() ? getApiKeyFromEnv() : getApiKeyFromFile();
+    if (freesoundApiKey == null || freesoundApiKey.isEmpty()) {
+        throw new IllegalStateException(
+                "Could not find Freesound API key value in environment or property file")
+    }
+    return "\"$freesoundApiKey\""
+}
+
+private boolean isCiBuild() {
+    System.getenv("TRAVIS") == "true"
```
peter-tackage: That mostly why I've stuck with using a file to define the `freesound.api.key` property for developers, rather moving to all environment variables - platform independence and no differences when using in command-line Gradle vs Android Studio.

PR #79, opened 2016-10-22
Don't end stream on search failure

### LINKED ISSUE #77: Search failure causes subsequent searches to not be performed
The following error causes the search activity's search capability to no longer function:

```
10-18 22:57:57.815 2738-2925/com.futurice.freesound.debug E/SearchViewModel: Error when setting search term
```

It would seem that the `onError` event is propagated too far, causing the searchQuery stream Observable to error and no longer accept requests.

This can occur when the search API call has a network failure. The app should instead be able to accept future queries.

peter-tackage:
Resolves #77 

I've added a set of tests which are marked `@Ignore` because they depend on that ability to remove the asynchronous nature of the Schedulers used in the `ActivityViewModel`. I figured this was the best way for now of defining the required behaviors. We should remove the `@Ignore` after #83

### DIFF
--- a/app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
+++ b/app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
@@ -79,7 +79,8 @@ final class DefaultSearchDataModel implements SearchDataModel {
     }
 
     @NonNull
-    private static Option<List<Sound>> toResults(final SoundSearchResult soundSearchResult) {
+    private static Option<List<Sound>> toResults(
+            @NonNull final SoundSearchResult soundSearchResult) {
         return Option.ofObj(soundSearchResult.results());
     }
 
--- a/app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
+++ b/app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
@@ -75,7 +75,7 @@ final class DefaultSearchDataModel implements SearchDataModel {
     @Override
     @NonNull
     public Completable clear() {
-        return Completable.fromAction(onClear());
+        return Completable.fromAction(clearResultAndError());
     }
 
     @NonNull
@@ -95,7 +95,7 @@ final class DefaultSearchDataModel implements SearchDataModel {
     }
 
     @NonNull
-    private Action onClear() {
+    private Action clearResultAndError() {
         return () -> {
             lastResultsStream.onNext(Option.none());
             lastErrorStream.onNext(Option.none());
--- a/app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
+++ b/app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
@@ -26,10 +26,13 @@ import java.util.List;
 
 import io.reactivex.Completable;
 import io.reactivex.Observable;
+import io.reactivex.functions.Action;
+import io.reactivex.functions.Consumer;
 import io.reactivex.subjects.BehaviorSubject;
 import polanski.option.Option;
 
 import static com.futurice.freesound.utils.Preconditions.get;
+import static timber.log.Timber.e;
 
 final class DefaultSearchDataModel implements SearchDataModel {
 
@@ -39,18 +42,22 @@ final class DefaultSearchDataModel implements SearchDataModel {
     @NonNull
     private final BehaviorSubject<Option<List<Sound>>> lastResultsStream = BehaviorSubject.create();
 
+    @NonNull
+    private final BehaviorSubject<Option<Throwable>> lastErrorStream = BehaviorSubject.create();
+
     DefaultSearchDataModel(@NonNull final FreeSoundSearchService freeSoundSearchService) {
         this.freeSoundSearchService = get(freeSoundSearchService);
     }
 
     @Override
-    @NonNull
     public Completable querySearch(@NonNull final String query) {
         return freeSoundSearchService.search(get(query))
                                      .map(SoundSearchResult::results)
                                      .map(Option::ofObj)
-                                     .doOnSuccess(lastResultsStream::onNext)
-                                     .toCompletable();
+                                     .doOnSuccess(storeValueAndClearError())
+                                     .doOnError(storeError(query))
+                                     .toCompletable()
+                                     .onErrorComplete();
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -61,7 +68,37 @@ public Completable querySearch(@NonNull final String query) {
...
+
+    @Override
+    @NonNull
     public Completable clear() {
-        return Completable.fromAction(() -> lastResultsStream.onNext(Option.none()));
+        return Completable.fromAction(clearResultAndError());
+    }
+
+    @NonNull
+    private Consumer<Throwable> storeError(final @NonNull String query) {
```
tomaszpolanski: Change order of `final @NonNull`

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -39,18 +42,22 @@
     @NonNull
     private final BehaviorSubject<Option<List<Sound>>> lastResultsStream = BehaviorSubject.create();
 
+    @NonNull
+    private final BehaviorSubject<Option<Throwable>> lastErrorStream = BehaviorSubject.create();
```
tomaszpolanski: Wouldn't it be now nicer to have this as OnceAndStream with default value?

app/src/test/java/com/futurice/freesound/feature/search/DefaultSearchDataModelTest.java
```diff
@@ -170,7 +166,79 @@ public void getSearchResults_emitsEmptyList_whenCleared() {
     }
 
     @Test
-    public void clear_emitsSingleUnit() {
+    public void getSearchErrorStream_hasNoDefaultResults() {
+        TestObserver<Option<Throwable>> ts = defaultSearchDataModel.getSearchErrorStream()
```
tomaszpolanski: Here you don't event need to assign `TestObserver` to a variable

app/src/test/java/com/futurice/freesound/feature/search/SearchViewModelTest.java
```diff
@@ -86,6 +88,24 @@ public void getSounds_emitsSearchResultsWrappedInDisplayableItems() {
         ts.assertValue(ofObj(expectedDisplayableItems(sounds)));
     }
 
+    @Test
+    public void search_recoversFromSearchErrors() {
+        ArrangeBuilder arrangeBuilder = new ArrangeBuilder().withErrorWhenSearching();
```
tomaszpolanski: Where's AAA?

peter-tackage: It's (AAA)^2 :-) I'll if I can do this neatly without adding to much to the ArrangeBuilder.

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -26,42 +26,81 @@
...
+                                     .doOnSuccess(storeValueAndClearError())
+                                     .doOnError(storeError(query))
+                                     .toCompletable()
+                                     .onErrorComplete();
+    }
+
+    @Override
+    @NonNull
+    public Observable<Option<List<Sound>>> getSearchResultsOnceAndStream() {
+        return lastResultsOnceAndStream.hide();
```
tomaszpolanski: Maybe instead of hide, we should use `observeOn`?

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -26,42 +26,81 @@
...
     }
 
     @Override
     @NonNull
-    public Observable<Option<List<Sound>>> getSearchResultsStream() {
-        return lastResultsStream.hide();
+    public Observable<Option<Throwable>> getSearchErrorOnceAndStream() {
+        return lastErrorOnceAndStream.hide();
```
tomaszpolanski: Maybe instead of hide, we should use `observeOn`?

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -26,42 +26,81 @@
...
     }
 
     @Override
     @NonNull
     public Completable clear() {
-        return Completable.fromAction(() -> lastResultsStream.onNext(Option.none()));
+        return Completable.fromAction(clearResultAndError());
```
tomaszpolanski: Why not `this::clearResultAndError`?

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -26,42 +26,81 @@
...
+    @NonNull
+    private Consumer<Option<List<Sound>>> storeValueAndClearError() {
+        return results -> {
+            lastResultsOnceAndStream.onNext(results);
+            lastErrorOnceAndStream.onNext(Option.none());
+        };
+    }
+
+    @NonNull
+    private Action clearResultAndError() {
```
tomaszpolanski: change to normal void method

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -26,42 +26,81 @@
...
 
     @Override
-    @NonNull
     public Completable querySearch(@NonNull final String query) {
         return freeSoundSearchService.search(get(query))
                                      .map(SoundSearchResult::results)
                                      .map(Option::ofObj)
-                                     .doOnSuccess(lastResultsStream::onNext)
-                                     .toCompletable();
+                                     .doOnSuccess(storeValueAndClearError())
```
tomaszpolanski: Change to `this::storeValueAndClearError`

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -26,42 +26,81 @@
...
+    @NonNull
+    private Consumer<Throwable> storeError(@NonNull final String query) {
+        return e -> {
+            lastErrorOnceAndStream.onNext(Option.ofObj(e));
+            e(e, "Error searching Freesound for query: %s ", query);
+        };
+    }
+
+    @NonNull
+    private Consumer<Option<List<Sound>>> storeValueAndClearError() {
```
tomaszpolanski: change to void method

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -26,42 +26,81 @@
...
     DefaultSearchDataModel(@NonNull final FreeSoundSearchService freeSoundSearchService) {
         this.freeSoundSearchService = get(freeSoundSearchService);
     }
 
     @Override
-    @NonNull
     public Completable querySearch(@NonNull final String query) {
         return freeSoundSearchService.search(get(query))
                                      .map(SoundSearchResult::results)
                                      .map(Option::ofObj)
```
tomaszpolanski: Remove this one, as it is not possible to ever have here null (would crash).
Move conversion to option to appropriate method

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
peter-tackage: I think I'd rather keep this Scheduler-neutral, if there's nothing downstream. The consumer of this Observable can then choose how they want to schedule it. If there was some downstream processing happening internally to the ViewModel, then I'd agree e.g. -
```
  return lastResultsOnceAndStream.observeOn(Schedulers.computation())
  .map(z -> z + 1) 
  .filter(y -> y > 1);
```

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
peter-tackage: See previous response

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -77,7 +75,17 @@ public Completable querySearch(@NonNull final String query) {
     @Override
     @NonNull
     public Completable clear() {
-        return Completable.fromAction(clearResultAndError());
+        return Completable.fromAction(this::clearResultAndError);
+    }
+
+    @NonNull
+    private static Option<List<Sound>> toResults(final SoundSearchResult soundSearchResult) {
```
tomaszpolanski: Nonnull

PR #93, opened 2016-11-26
[CHORE] Tidy data binding

peter-tackage:
Renaming some things to avoid clashing with all the other things that use that name.


app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
```diff
@@ -70,7 +70,7 @@
     private final AtomicOption<Unbinder> unbinder = new AtomicOption<>();
 
     @NonNull
-    private final Binder binder = new Binder() {
+    private final DataBinder dataBinder = new DataBinder() {
```
tomaszpolanski: Replace with `SimpleDataBinder` just like in SearchActivity

PR #95, opened 2016-12-02
Use Auto factory

peter-tackage:


app/src/main/java/com/futurice/freesound/network/api/model/Sound.java
```diff
@@ -110,4 +107,49 @@
         return new AutoValue_Sound.Builder();
     }
 
+    @AutoValue
+    public abstract static class Image implements Parcelable {
```
peter-tackage: This change is required because the Gson extension for AutoValue complains about using a `Map` with an non-`String` key. Additionally, I decided that I didn't like exposing the enum anyway.

peter-tackage:
This has been at review for a while, so I'm just going to merge these, if anyone disagrees then we can address it later.

PR #100, opened 2017-01-06
Add basic duration label

peter-tackage:


app/src/main/java/com/futurice/freesound/feature/common/waveform/PlaybackWaveformView.java
```diff
@@ -0,0 +1,81 @@
...
+        ButterKnife.bind(this);
+    }
+
+    @Override
+    public void setWaveform(@NonNull final float[] waveform) {
+        waveformView.setWaveform(waveform);
+    }
+
+    @Override
+    public void clear() {
```
peter-tackage: Consider making the `clear` and `set` only operate on the async waveform.

PR #101, opened 2017-01-08
Add progress observable and observableExoPlayer

peter-tackage:
This is only the definition of the observable, it's not wired up to any functionality (just yet).

### DIFF
--- a/app/src/main/java/com/futurice/freesound/feature/audio/ObservableExoPlayer.java
+++ b/app/src/main/java/com/futurice/freesound/feature/audio/ObservableExoPlayer.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2017 Futurice GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/futurice/freesound/feature/audio/ObservableExoPlayer.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.reactivex.Observable;
+
+/**
+ * Represent the {@link com.google.android.exoplayer2.ExoPlayer} through Observables.
+ */
+public interface ObservableExoPlayer {
+
+    @NonNull
+    Observable<ExoPlayerState> getExoPlayerStateStream();
```
tomaszpolanski: Do we document interfaces?

peter-tackage: I guess so, it's a bit tedious though.

PR #106, opened 2017-01-15
Enable Jacocco coverage in debug

peter-tackage:


peter-tackage:
I think this only covers either unit tests or instrumentation tests at the moment. Even incomplete, it doesn't hurt to merge it.

PR #107, opened 2017-02-04
Show Search Progress indicator

maritoast:
Solution with a new Stream, alternatively the error stream could also be transformed into a state stream instead, what do you think?


maritoast:
I could also introduce a little bit of databinding if you would like

peter-tackage:
Thanks! You bring up an excellent point about using multiple streams vs a single state value. Having multiple streams may be an unnecessary complication. The trouble is then, how do you model all the required data for InProgress,Error and Success in a single data type. This could be done with `Option` but even that is a bit painful because that is really only designed for a `Some`/`None` evaluation. 

Kotlin has a notion of [Sealed Classes](https://kotlinlang.org/docs/reference/classes.html#sealed-classes) which essentially allows you to define a class as having a single values from an enumerated range of other classes.

[RxSealedUnions](https://github.com/pakoito/RxSealedUnions) is an implementation for Java 6  which might be useful. I haven't tried it personally though.

We could make use of either Kotlin or RxSealedUnions to define a single class with the appropriate states and expose it as a single stream for the View to consume.

Regarding Android Data Binding, I'd prefer not to use them here because we have a solution for view binding already, so I'd prefer to keep it consistent for now.

apoi:
If I understand the problem here correctly, the approach in Reark is to wrap the value/error in a [DataStreamNofitication](https://github.com/reark/reark/blob/master/reark/src/main/java/io/reark/reark/data/DataStreamNotification.java) object. Progress start, error, completion with value, or completion without value are all signaled with the notification object. It's pretty clean way to keep the state in a single stream.

A viewmodel then subscribes to a stream that emits notifications, and filters based on the event type. Reark example for this is in [RepositoriesViewModel](https://github.com/reark/reark/blob/master/appshared/src/main/java/io/reark/rxgithubapp/shared/viewmodels/RepositoriesViewModel.java#L145).

peter-tackage:
Yep, it's the same problem. The Reark approach look pretty good, although I'd prefer if we could avoid using nullability. I'll have a look at the RxSealedUnions library to see if we can make use of that. Reark has some great stuff in it and I'd like to leverage more if it into parts of Freesound.

app/src/main/java/com/futurice/freesound/feature/search/SearchState.java
```diff
@@ -0,0 +1,69 @@
...
+        this.value = false;
+        this.optionalThrowable = Option.ofObj(throwable);
+    }
+
+    @NonNull
+    Option<Throwable> getError() {
+        return optionalThrowable;
+    }
+
+    boolean searchTriggered() {
```
peter-tackage: `inProgress`

app/src/main/java/com/futurice/freesound/feature/search/SearchState.java
```diff
@@ -0,0 +1,69 @@
...
+        return new SearchState(false);
+    }
+
+    @NonNull
+    static SearchState SEARCH_ERROR(@NonNull Throwable throwable) {
+        get(throwable);
+        return new SearchState(throwable);
+    }
+
+    private final boolean value;
```
peter-tackage: isInProgress

app/src/main/java/com/futurice/freesound/feature/search/SearchState.java
```diff
@@ -0,0 +1,69 @@
...
+import android.support.annotation.NonNull;
+
+import polanski.option.Option;
+
+import static com.futurice.freesound.common.utils.Preconditions.get;
+
+final class SearchState {
+
+    @NonNull
+    static SearchState SEARCH_TRIGGERED() {
```
peter-tackage: Just use `isInProgress`, `idle` and `errored` here

app/src/main/java/com/futurice/freesound/feature/search/SearchState.java
```diff
@@ -0,0 +1,69 @@
...
+    Option<Throwable> getError() {
+        return optionalThrowable;
+    }
+
+    boolean searchTriggered() {
+        return value;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
peter-tackage: I don't think this is necessary - it gets autogenerated. (Also note that if you implement your own `equals()` you would need to also implement your own `hashCode()`

app/src/main/res/layout/fragment_search.xml
```diff
@@ -15,23 +15,34 @@
...
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:textSize="@dimen/text_title"
+            android:visibility="gone"
+            android:text="@string/search_no_results_found"
+            android:layout_below="@+id/progressBar_searchProgress"/>
 
-</FrameLayout>
\ No newline at end of file
+</RelativeLayout>
```
peter-tackage: Add a newline at the end.

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -93,19 +97,19 @@ public Completable clear() {
 
     private void storeValueAndClearError(@NonNull final Option<List<Sound>> listOption) {
         lastResultsOnceAndStream.onNext(listOption);
-        lastErrorOnceAndStream.onNext(Option.none());
+        searchStateOnceAndStream.onNext(SEARCH_CLEARED());
```
peter-tackage: I think this reads much better than before - it's more obvious what the event is.

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -59,6 +62,7 @@
 
     @Override
     public Completable querySearch(@NonNull final String query) {
+        searchStateOnceAndStream.onNext(SEARCH_TRIGGERED());
```
peter-tackage: Make this a side-effect of subscribing to the returned `Completable` - otherwise progress will begin as soon as this is called, not when the actual action is performed.

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -59,6 +62,7 @@
...
         return freeSoundSearchService.search(get(query))
                                      .map(DefaultSearchDataModel::toResults)
                                      .doOnSuccess(this::storeValueAndClearError)
```
peter-tackage: You also need to clear progress whenever the `Completable` terminates - either success or error.

app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.java
```diff
@@ -170,4 +181,12 @@ private void showResults(@NonNull final List<DisplayableItem> sounds) {
         }
     }
 
+    private void showProgress(@NonNull final SearchState searchState) {
+        if (searchState.searchTriggered()) {
+            get(progressBar).setVisibility(View.VISIBLE);
```
peter-tackage: You could write this as:
```
get(progressBar).setVisibility(searchState.searchTriggered() ? View.VISIBLE : View.GONE);
```

Although there's probably value in writing utility class with `setVisibility(View view, boolean isVisible)` and using that.

app/src/main/res/layout/fragment_search.xml
```diff
@@ -15,23 +15,34 @@
...
-             android:layout_width="match_parent"
-             android:layout_height="match_parent">
+<RelativeLayout
+        xmlns:android="http://schemas.android.com/apk/res/android"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent">
+
+    <ProgressBar
+            style="@android:style/Widget.ProgressBar.Horizontal"
+            android:id="@+id/progressBar_searchProgress"
```
peter-tackage: Keep the `android:id` first. Then apply common View attributes, then attributes only applicable to this ProgressBar. Do this generally in this layout.

app/src/test/java/com/futurice/freesound/feature/search/DefaultSearchDataModelTest.java
```diff
@@ -127,10 +128,12 @@ public void getSearchResultsOnceAndStream_hasNoTerminalEvent() {
...
     @Test
     public void getSearchErrorOnceAndStream_hasNoneAsDefault() {
-        defaultSearchDataModel.getSearchErrorOnceAndStream()
-                              .test()
-                              .assertNotTerminated()
-                              .assertValue(Option.none());
+        assertEquals(defaultSearchDataModel.getSearchStateOnceAndStream()
+                                           .test()
+                                           .assertNotTerminated()
+                                           .values()
```
peter-tackage: Split this up into different assertions (don't nest them). Also use assertJ assertions, rather than `junit.framework.Assert.*`

app/src/test/java/com/futurice/freesound/feature/search/DefaultSearchDataModelTest.java
```diff
@@ -140,9 +143,11 @@ public void getSearchErrorOnceAndStream_emitsErrorValue_whenQuerySearchErrors()
                          .act()
                          .querySearch();
 
-        defaultSearchDataModel.getSearchErrorOnceAndStream()
-                              .test()
-                              .assertValue(Option.ofObj(searchError));
+        assertEquals(defaultSearchDataModel.getSearchStateOnceAndStream()
```
peter-tackage: Same comments as the other test.

app/src/main/java/com/futurice/freesound/feature/search/SearchState.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.futurice.freesound.common.utils.Preconditions;
+
+import android.support.annotation.NonNull;
+
+import polanski.option.Option;
+
+import static com.futurice.freesound.common.utils.Preconditions.get;
+
+final class SearchState {
```
peter-tackage: This should be an AutoValue class - that's how you get equals hashCode etc for free.

peter-tackage: I'm unsure if this class should also contain the result values. let's discuss tomorrow.

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
peter-tackage: You will need a test for that too.

peter-tackage:
@maritoast there's a compilation error.

maritoast:
@peter-tackage Can i trigger the build again somehow? It compiles and executes for me and the tests all pass

peter-tackage:
@maritoast https://travis-ci.org/futurice/freesound-android/builds/204611845 "Restart Build" button on right.

peter-tackage:
Make sure that you've push the latest too. You probably need to rebase as well.

maritoast:
there we go

app/src/test/java/com/futurice/freesound/feature/search/SearchFragmentViewModelTest.java
```diff
@@ -124,21 +124,24 @@ public void stopPlayback_stopsAudioPlayback() {
...
-            when(searchDataModel.getSearchResultsOnceAndStream())
+            when(searchDataModel.getSearchStateOnceAndStream())
                     .thenReturn(mockedSearchResultsStream);
             return this;
         }
 
         Arrangement enqueueSearchResults(@NonNull final Option<List<Sound>> sounds) {
-            mockedSearchResultsStream.onNext(sounds);
+            sounds
+                    .ifSome(soundList -> mockedSearchResultsStream
```
tomaszpolanski: Move `ifSome` to the same line as `sounds`

VasileUngureanu:
Hi, 
Maybe these notices are not important for this project showcase but anyway: 

1. Search delay (after changing text) is too long.
2. "In progress" animation is not smooth, it animate like there is a "lag" and make impression of bad performance to all screen. 
3. "No results found" label can be positioned on better place.

PR #108, opened 2017-02-04
Use scheduler provider

peter-tackage:
I thought I'd follow the advice of my own blog and use a `SchedulerProvider`


app/src/main/java/com/futurice/freesound/feature/common/scheduling/DefaultSchedulerProvider.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @NonNull
+    @Override
+    public Scheduler io() {
+        return Schedulers.io();
+    }
+
+    @NonNull
+    @Override
+    public Scheduler time(@NonNull String tag) {
```
tomaszpolanski: final

PR #113, opened 2017-02-09
Add Freesound Auth API

peter-tackage:
Note: This has a breaking change to the `freesound-api.properties` file -  a property for clientId has been added the key property has been renamed to clientSecret for consistency.


apoi:
This just adds the API but doesn't use it for anything yet, right?

peter-tackage:
Correct @apoi, no functional changes just yet.

PR #114, opened 2017-02-09
#15 Add generic/reusable logic for recycler adapter

luciapayo:


app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/DiffUtilCallback.java
```diff
@@ -0,0 +1,52 @@
+package com.futurice.freesound.feature.common.ui.adapter;
+
+import com.futurice.freesound.feature.common.DisplayableItem;
+
+import android.support.annotation.NonNull;
+import android.support.v7.util.DiffUtil;
+
+import java.util.List;
+
+final class DiffUtilCallback extends DiffUtil.Callback {
```
peter-tackage: Don't forget to add the standard copyright headers to the class files. Unfortunately, Apache v2.0 requires this. You should be able to achieve this automatically through the style files.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/DiffUtilCallback.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @NonNull
+    private final List<DisplayableItem> mOldItems;
```
peter-tackage: We don't use the `m` prefix for fields.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Implementation of {@link android.support.v7.widget.RecyclerView.Adapter} for {@link
+ * DisplayableItem}.
+ */
+@FragmentScope
+public final class RecyclerViewAdapter extends RecyclerView.Adapter {
+
+    @NonNull
+    private final List<DisplayableItem> mItems = new ArrayList<>();
```
peter-tackage: No `m` prefix.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+    @NonNull
+    private final IListItemComparator mComparator;
+
+    @NonNull
+    private final IViewHolderFactory mInstantiator;
+
+    @NonNull
+    private final IViewHolderBinder<DisplayableItem> mBinder;
+
+    private Executor mExecutor = Executors.newSingleThreadExecutor();
```
peter-tackage: `final` and `@NonNull`

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+    public int getItemViewType(final int position) {
+        return mItems.get(position).type().ordinal();
+    }
+
+    /**
+     * Updates mItems currently stored in adapter with the new mItems.
+     *
+     * @param items collection to update the previous values
+     */
+    public void update(@NonNull final List<DisplayableItem> items) {
```
peter-tackage: For public API, I think it's work enforcing the `@NonNull` condition,

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private void updateItems(@NonNull final List<DisplayableItem> items) {
+        AndroidPreconditions.assertWorkerThread();
+
+        mItems.clear();
+        mItems.addAll(items);
+    }
+
+    private void updateAdapterWithDiffResult(@NonNull final DiffUtil.DiffResult result) {
```
peter-tackage: could be static

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
peter-tackage: Can you explain to me the reasons behind the use of the single thread executor? Does this need to be pushed to another (single) thread?

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
peter-tackage: (here and other fields)

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/IListItemComparator.java
```diff
@@ -0,0 +1,26 @@
+package com.futurice.freesound.feature.common.ui.adapter;
+
+import com.futurice.freesound.feature.common.DisplayableItem;
+
+public interface IListItemComparator {
```
peter-tackage: The convention in this project is to not use the `I` prefix for interfaces. (This is not something that I'm 100% convinced of ... but those are the rules :-) )

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+    @NonNull
+    private final IViewHolderBinder<DisplayableItem> mBinder;
+
+    private Executor mExecutor = Executors.newSingleThreadExecutor();
+
+    @Inject
+    RecyclerViewAdapter(@NonNull final IListItemComparator comparator,
+                        @NonNull final IViewHolderFactory instantiator,
+                        @NonNull final IViewHolderBinder<DisplayableItem> binder) {
+        mComparator = comparator;
```
peter-tackage: Although other classes enfore `@NonNull` - you are correct to not do it here. We've decided that it's unnecessary boilerplate to use `get()` when Dagger typically enforced non-null dependencies here. So either intentionally or otherwise 👍

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+        mInstantiator = instantiator;
+        mBinder = binder;
+    }
+
+    @Override
+    public RecyclerView.ViewHolder onCreateViewHolder(final ViewGroup parent, final int viewType) {
```
peter-tackage: For public API it's worth declaring and enforcing `@NonNull` for the parameters - this will make it more straightforward if/when we extract it to separate module.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+    @NonNull
+    private final IViewHolderFactory mInstantiator;
+
+    @NonNull
+    private final IViewHolderBinder<DisplayableItem> mBinder;
+
+    private Executor mExecutor = Executors.newSingleThreadExecutor();
+
+    @Inject
+    RecyclerViewAdapter(@NonNull final IListItemComparator comparator,
```
peter-tackage: No `I` prefix for interfaces... unless you can come up with an particularly convincing, project-wide argument. (I'd love to find a better way of naming our "default" implementations)

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+import io.reactivex.Observable;
+import io.reactivex.android.schedulers.AndroidSchedulers;
+import io.reactivex.schedulers.Schedulers;
+
+/**
+ * Implementation of {@link android.support.v7.widget.RecyclerView.Adapter} for {@link
+ * DisplayableItem}.
+ */
+@FragmentScope
+public final class RecyclerViewAdapter extends RecyclerView.Adapter {
```
peter-tackage: I'm a little bit skeptical about marking this as `@FragmentScope`. Generally speaking, we shouldn't enforcing this here, but in the usage as required by the application - so leave this to the Dagger module (this is again based upon the idea that this adapter logic should not depend on how it is used in this particular application).

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/IListItemComparator.java
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * Decides whether two {@link DisplayableItem} represent the same Item.
+     * For example, if your items have unique ids, this method should check their id equality.
+     *
+     * @return True if the two items represent the same object or false if they are different.
+     */
+    boolean areItemsTheSame(final DisplayableItem item1, final DisplayableItem item2);
+
+    /**
+     * Checks whether the visual representation of two {@link DisplayableItem}s is the same.
```
peter-tackage: "are the same"

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/IListItemComparator.java
```diff
@@ -0,0 +1,26 @@
...
+     *
+     * This method is called only if {@link #areItemsTheSame(DisplayableItem, DisplayableItem)}
```
peter-tackage: I'm not sure you can usefully declaring how/when this method is called. I think you're better off just declaring it's purpose and leave any comments about *when* this method is called to points where this is called.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/DiffUtilCallback.java
peter-tackage: (just don't forget to ensure that it's not applied for your 9-5 work!)

peter-tackage:
In addition to my other comments; can we make this JVM testable/tested. The primary dependencies are the support library, so it should be achievable.

luciapayo:
Welcome back @peter-tackage! :)

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/DiffUtilCallback.java
luciapayo: I can't figure out how to import the copyright xml config files. I imported the style xml but that is not enough, right?

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/IListItemComparator.java
luciapayo: This interface is a way to adapt the DiffUtils Callback to DisplayableItem. I took the documentation from the original platform's Callback and tried to adapt it. This part you refer to is in the original documentation and I found it useful when I was trying to understand how the DiffUtils works, that's why I decided to include it.
This is called by the DiffUtils (through the DiffUtilCallback class I implemented), I don't think there is a better place to include this documentation.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
luciapayo: The idea was to calculate the difference of the lists in the background, which required reading/writing the items in the adapter in a single background thread to ensure the integrity of the data without actual synchronisation (slow). The problem is that the framework does some calls that require reading of the items in the adapter, getItemCount for instance, and this calls happen in the main thread causing potential race conditions. 
Let's enforce main thread for now since calculating the difference in a background thread is not that easy.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
luciapayo: Yes, you are right, this shouldn't be FragmentScope

luciapayo:
I've reworked the adapter a bit based on the experience we got using it in N26:

- Calculate diff in main thread to avoid race conditions (explained in one of the comments above)
- Do not update adapter through DiffResult when it is empty, it uses notifyItemRange which is much slower than notifyDataSetChanged when all items have changed. 
- Use integer instead of enum for the item types. There are two reasons for this:
    * The `RecyclerView.Adapter` uses integers, we would need to transform the enum to integer and back with the resulting boilerplate.
    * Dagger multi binding doesn't support enums. Also it is straight forward in the adapter to get the right factory or binder from the type passed if we use integers.
- The Binder and the Factory are inner classes of the ViewHolder. This I think is more a personal preference. The Factory and the Binder are boilerplate to get the generic Adapter running so I think it is a bit overkill to create a file for each.

luciapayo:
Also I'd like to challenge the idea of binding/unbind the view holders to view models. I think in the case of the SoundItems there is only two views that need to be bound, let's break it down:
- For title, description, duration and thumbnail: This is more static info not likely to change often, but even when it changes, it can go through the normal adapter's update mechanism. So create a new list for the whole set and do `adapter.update(newList)`, the DiffUtils will take care of calculating what has changed and will update only those items. Otherwise the whole DiffUtils mechanism is pretty useless.
- For the WaveView and progressPercentage: This views change a lot as the sound plays and I get why a ViewModel is needed here BUT I think the view model should be bound to the view itself and not to the view holder. The view has lifecycle that can be used for bind/unbind (there is a very nice post about this topic but I can't find it right now, maybe you also know it?), that way we can remove a lot of complexity and boiler plate in the BaseBindingViewHolder and its childs.

I don't think the adapters in Android are designed to bind each element to a view model. Also, it is weird that some updates go through the fragment's view model and others go through the view holder's view model, I don't think this is consistent.

luciapayo:
The tests are missing, will include them later

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/DiffUtilCallback.java
peter-tackage: I'm not 100% certain of how to do it, I would have thought that importing the top level settings would be enough.

If you need to import the Copyright directly then: Preferences -> Editor -> Copyright -> Copyright Profiles

You can then import the file from the repo: <repo>/ide/copyright/

And I think you need to then set that as your copyright profile for this picture.

peter-tackage: I'm not 100% certain of how to do it, I would have thought that importing the top level settings would be enough.

If you need to import the Copyright directly then: Preferences -> Editor -> Copyright -> Copyright Profiles

You can then import the file from the repo: <repo>/ide/copyright/

And I think you need to then set that as your copyright profile for this project.

peter-tackage:
Good point about the ViewItem binding vs DiffUtils. It's consistent to allow for updating those static model values outside of the DiffUtils mechanism. We don't always want to go through the DiffUtils, but the question is how to approach that in a clearly defined way.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/DiffUtilCallback.java
luciapayo: Done

app/src/testDebug/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapterTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+        adapter.update(null);
+    }
+
+    @Test
+    public void diffUtilsIsNotUsedWhenUpdatingAdapterForTheFirstTime() {
+        List<DisplayableItem> itemList = itemList();
+
+        // FIXME: AndroidPreconditions.assertUiThread(), how to mock this ?
+        adapter.update(itemList);
```
luciapayo: `AndroidPreconditions.assertUiThread()` is called in update method but I don't know how to mock it or make it work. I commented that line to run the tests and check if they pass but we should figure out how to work around this, any ideas?

app/src/testDebug/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapterTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+        verify(binder2).bind(viewHolder, itemList.get(1));
+    }
+
+    @Test
+    public void unbindIsCalledForBindingViewHolderInOnViewRecycled() {
+        BaseBindingViewHolder viewHolder = mock(BaseBindingViewHolder.class);
+
+        //FIXME: This throws NPE
+        adapter.onViewRecycled(viewHolder);
```
luciapayo: The unbind() method in the BindingViewHolder throws a NPE because it is a mock. I need it to be a mock to be able to verify that the unbind() method is called. Any ideas how to test this?

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    @Override
+    public void onBindViewHolder(final RecyclerView.ViewHolder holder, final int position) {
+        final DisplayableItem item = modelItems.get(position);
+        binderMap.get(item.type()).bind(get(holder), item);
+    }
+
+    @Override
+    public void onViewRecycled(final RecyclerView.ViewHolder holder) {
```
luciapayo: I don't like the fact that the adapter knows about the BaseBindingViewHolder class, it should be generic, but it is the easiest way to work around the unbinding. I think it would be nice to remove the binding/unbinding of view models in the adapter, I have some ideas, probably I'll open another PR soon.

luciapayo:
@peter-tackage I've included tests for the recycler adapter. I had some problems and I need a bit of help, check the comments out.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
peter-tackage: If there's something like, just do what you suggested: make the changes in another PR. Otherwise we'll have PRs open forever trying to find the perfect solution.

app/src/testDebug/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapterTest.java
peter-tackage: The `SchedulerProvider` has a method `isUiThread` which mockable. The assertion in the Adapter could be changed to use that, rather than `AndroidPreconditions`.

app/src/testDebug/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapterTest.java
luciapayo: I have removed this test for now since I don't know how to solve the problem

luciapayo:
@peter-tackage ready for final review :)

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/DiffUtilCallback.java
```diff
@@ -0,0 +1,67 @@
...
+    @NonNull
+    private final List<DisplayableItem> newItems;
+
+    @NonNull
+    private final ItemComparator comparator;
+
+    DiffUtilCallback(@NonNull final List<DisplayableItem> oldItems,
+                     @NonNull final List<DisplayableItem> newItems,
+                     @NonNull final ItemComparator comparator) {
+        this.oldItems = oldItems;
```
peter-tackage: I'm probably contradicting myself here, but I think this should have null checks on the values. I say this because the values are calculated, rather than being simply and reliable created via Dagger.

peter-tackage:
Hey thanks @luciapayo. I'll take a look.

On thing I've noticed is that I seem to get two emissions when changing the search term after receiving results:
1. Search for "cats" -> get some sounds
2. Remove the existing search term using the clear button -> sounds are cleared
3. Search for "dogs" -> results for "cats" quickly appear before the results for "dogs" are shown

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,158 @@
...
+
+        if (modelItems.isEmpty()) {
+            updateAllItems(items);
+        } else {
+            updateDiffItemsOnly(items);
+        }
+    }
+
+    /**
+     * Only use for the first update of the adapter, whe it is still empty.
```
peter-tackage: "when"

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,158 @@
...
+    private void updateItemsInModel(@NonNull final List<DisplayableItem> items) {
+        modelItems.clear();
+        modelItems.addAll(items);
+    }
+
+    private void updateAdapterWithDiffResult(@NonNull final DiffUtil.DiffResult result) {
+        result.dispatchUpdatesTo(this);
+    }
+
+    private void assertUiThread() {
```
peter-tackage: What do you think about having this method in the `SchedulerProvider`? I can see it being useful for reuse.

luciapayo:
@peter-tackage I've included your suggestions.

About the extrange behavior that you described with the search results, I've noticed that is related to hiding/showing the recyclerview when the results are empty because in this scenario the recycler view is just hidden but not updated. This problem is similar to the one I wrote my blog post about, instead of updating the whole state of the recycler view (visibility + items inside + etc.) it is just sending the command to hide/show, causing it to not be aligned with the actual updating of the items inside, giving the impression that there are two emissions the next time is updated:
- Search for cats -> Cats are put in the recycler view
- Delete search term -> recycler view is hidden but cats are still inside
- Search for dogs -> Show recycler view (has cats), then update the recycler adapter (async. Diff utils might take time) and finally show dogs.

Remember, always states of the view, never commands ;)

How I could see this clearly was commenting some lines in the fragment:
screen shot 2017-04-15 at 10 57 52

Also, I tested with master and the behavior is the same, it just happens faster.

In any case I think this is a separate problem not related to the implementation of the recycler adapter and should be fixed in a different PR.

peter-tackage:
Yeah, we originally had a more singular approach to updates and I've started work on another PR that will essentially remove the use of the "command-like" updates that got added during that last PR I merged. Which by the looks of it, was a terrible idea.

app/src/main/java/com/futurice/freesound/feature/search/SearchResultItemComparator.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.futurice.freesound.feature.search;
+
+import com.futurice.freesound.feature.common.DisplayableItem;
+import com.futurice.freesound.feature.common.ui.adapter.ItemComparator;
+
+public class SearchResultItemComparator implements ItemComparator {
```
peter-tackage: - Can be package visible, rather than public.
- Make class final

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewHolder.java
```diff
@@ -121,4 +129,46 @@ protected DataBinder getViewDataBinder() {
...
+        private final SoundItemViewModelFactory viewModelFactory;
+
+        SoundItemViewHolderBinder(@NonNull final SoundItemViewModelFactory viewModelFactory) {
+            this.viewModelFactory = viewModelFactory;
+        }
+
+        @Override
+        public void bind(@NonNull final RecyclerView.ViewHolder viewHolder,
+                         @NonNull final DisplayableItem item) {
+            SoundItemViewHolder soundItemViewHolder = SoundItemViewHolder.class.cast(viewHolder);
```
peter-tackage: TIL - I have never seen `cast` before!

peter-tackage: Any benefit over using `Sound sound = (Sound)item.model();`?

peter-tackage:
@luciapayo I've made my review and I've just pushed my own small changes for some minor stuff. If you're happy with that, then merge.

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewHolder.java
luciapayo: Not that I know of, just a different way for casting objects that I personally find more natural/readable :)

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -0,0 +1,151 @@
...
+    @Override
+    public void onBindViewHolder(final RecyclerView.ViewHolder holder, final int position) {
+        final DisplayableItem item = modelItems.get(position);
+        binderMap.get(item.type()).bind(get(holder), item);
+    }
+
+    @Override
+    public void onViewRecycled(final RecyclerView.ViewHolder holder) {
+        super.onViewRecycled(holder);
+        if (holder instanceof BaseBindingViewHolder) {
```
peter-tackage: This `instanceof` seems a little out of place here. Ideally we shouldn't know about those details here. Was this one of the things on your next PR "todo" list?

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -121,7 +121,7 @@ private void updateAllItems(@NonNull final List<DisplayableItem> items) {
     }
 
     /**
-     * Do not use for first update of the adapter.
+     * Only used for first update of the adapter.
```
luciapayo: This is wrong since the idea is that this method is never used for the first update. Was the previous wording incorrect? 

> Do not use for first update of the adapter

app/src/main/java/com/futurice/freesound/feature/search/SearchFragmentViewModel.java
```diff
@@ -74,9 +75,8 @@ void openSoundDetails(@NonNull final Sound sound) {
 
     @NonNull
     private static List<DisplayableItem> wrapInDisplayableItem(@NonNull final List<Sound> sounds) {
-        return Observable.fromIterable(sounds)
-                         .map(sound -> DisplayableItem.create(sound, SOUND))
-                         .toList()
-                         .blockingGet();
+        return Ix.from(sounds)
```
luciapayo: I didn't know about this! looks more convenient, I'll start using it

luciapayo:
@peter-tackage there's is a minor thing with one of the comments you corrected in the code, we can roll back to the original comment or figure another way to reword it.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
luciapayo: Yes, exactly, I totally agree this is out of place. And yes, this is what we talked about in a previous comment, I am planning to open a new PR to tackle the bind/unbind case.

app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
peter-tackage: Gah, I'll fix it. My intent in changing it was to word in a way that didn't imply that it was part of API, but just the mechanism - so to describe more how it was being used. I'll correct it.

app/src/main/java/com/futurice/freesound/feature/search/SearchFragmentViewModel.java
peter-tackage: It may become unnecessary once Java 8 streams are used, but it's generally faster than RxJava2 for synchronous stuff.

luciapayo:
Alright, looks good! I'll merge as soon as Travis is done

PR #120, opened 2017-03-31
SonarQube improvements

peter-tackage:


peter-tackage:
This is no longer necessary, I was attempting to have SonarQube run on each PR, but this is actually achieved through the Sonar Github plugin and can also be done locally via SonarLint plugin for Android Studio.

PR #124, opened 2017-04-01
Search Progress UI improvements

peter-tackage:
I'm making some small improvements to the UI for search progress reporting.

1. Report as in progress when the user starts typing
2. Change styling to be material design based
3. Center the "no results text"
4. Add OK dismiss to Snackbar
5. Change color of progress indicator

Because I can't help myself, I'm also fixing updating our Gradle plugin and fixing Travis for the latest support lib.


app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -54,21 +54,22 @@
...
-        return freeSoundApiService.search(get(query))
-                                  .map(DefaultSearchDataModel::toResults)
-                                  .doOnSubscribe(__ -> searchStateOnceAndStream
-                                          .onNext(SearchState.inProgress()))
-                                  .doOnSuccess(this::storeValueAndClearError)
-                                  .doOnError(storeError(query))
-                                  .toCompletable()
-                                  .onErrorComplete();
+    public Completable querySearch(@NonNull final String query,
+                                   @NonNull final Completable afterThis) {
```
maritoast: Could you rename this to maybe preliminary or something similar? afterThis seems a bit weird in the chain below.

app/src/main/java/com/futurice/freesound/feature/search/SearchDataModel.java
```diff
@@ -24,7 +24,7 @@
 interface SearchDataModel {
 
     @NonNull
-    Completable querySearch(@NonNull String query);
+    Completable querySearch(@NonNull String query, @NonNull final Completable afterThis);
```
maritoast: and here

peter-tackage: I was never very happy with the name I gave this...

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
```diff
@@ -54,21 +54,22 @@
...
+        return afterThis.doOnSubscribe(__ -> reportInProgress())
```
maritoast: is there an easy way here to send a new state other than idle with the last result? Otherwise this will clear the previous search result and thus the listview instantly. Alternatively remember the last result in the SearchFragment until there are no searchresults

app/src/main/res/layout/fragment_search.xml
```diff
@@ -15,34 +15,35 @@
...
+            android:visibility="gone"
+            android:layout_marginTop="-4dp"
+            style="@style/ProgressBarStyle"/>
+
     <TextView
             android:id="@+id/textView_searchNoResults"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:layout_below="@+id/progressBar_searchProgress"
             android:layout_gravity="center"
```
maritoast: Maybe use the chance to change the gravity here, as one never sees that the query failed with the keyboard open. Another way would be to make the layout consider the keyboard as the bottom anchor if it's active (Not sure how easy this is though)

maritoast:
I noticed the error snackbar gets hidden under the keyboard again, but I dont think any of the changes here caused that, I think my solution was to put ```android:windowSoftInputMode="adjustResize"``` 
into the manifest, but I think you removed it there some other time. Did that cause undesired behaviour?

app/src/main/java/com/futurice/freesound/feature/search/DefaultSearchDataModel.java
peter-tackage: This is possible with some work. It should be done in the `SearchDataModel`, rather than the View so that the SearchState emissions are the source of truth.

PR #128, opened 2017-04-14
Improve sound card

peter-tackage:
Adding some more details to the sound card


peter-tackage:
Review timeout - so merging.

PR #129, opened 2017-04-15
Fix compiler warnings

joan-domingo:


app/build.gradle
```diff
@@ -207,4 +207,12 @@ dependencies {
 
 }
 
+allprojects {
```
peter-tackage: Would it make more sense to have this in the top-level Gradle file?

peter-tackage: We only have one project at the moment, but I imagine that we would want to apply these checks globally.

joan-domingo: Indeed it would make more sense! I did not think it through

joan-domingo:
I moved the gradle task to the top-level gradle file. Let me know if you have any more questions.

peter-tackage:
Squash and merge when you are ready. Congrats on your first PR 🍾

joan-domingo:
wohoo 🎆

PR #133, opened 2017-06-28
Use Dagger subcomponents

peter-tackage:
Use Dagger subcomponents to reduce the boilerplate when Fragments requires singleton dependencies.


app/src/main/java/com/futurice/freesound/feature/search/SearchActivity.java
```diff
@@ -102,6 +102,23 @@ public void bind(@NonNull final CompositeDisposable d) {
                                  .subscribe(SearchActivity.this::handleErrorState,
                                             e -> e(e, "Error receiving Errors")));
         }
+
+        private void subscribeToSearchView(@NonNull final SearchView view,
```
tomaszpolanski: Why is this in this PR?

peter-tackage: It was a suggestion by the IDE that being flagged as a warning. I thought I'd clean it up while I was there (it's just moving the method).

PR #134, opened 2017-06-28
Add Kotlin

peter-tackage:
Add Kotlin and convert a view model. I have no immediate plans to convert all of the app, just to use Kotlin where it could be useful.


app/build.gradle
```diff
@@ -118,6 +120,9 @@ ext {
 
 dependencies {
 
+    // Kotlin
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jre7:$dependencyVersions.kotlin"
```
joan-domingo: I think it would make sense to use `org.jetbrains.kotlin:kotlin-stdlib-jre8` if we are targeting jdk 8

peter-tackage: I believe that the Kotlin JRE8 is not suitable for the Android runtime.

PR #137, opened 2017-07-03
Fixing a bit DisplayableItem generics

tomaszpolanski:
still, there is plenty of work to do


app/src/main/java/com/futurice/freesound/feature/common/ui/adapter/RecyclerViewAdapter.java
```diff
@@ -38,10 +38,10 @@
 /**
  * Implementation of {@link RecyclerView.Adapter} for {@link DisplayableItem}.
  */
-public final class RecyclerViewAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder> {
+public final class RecyclerViewAdapter<T> extends RecyclerView.Adapter<RecyclerView.ViewHolder> {
```
luciapayo: The whole idea of DisplayableItem is to reduce boiler plate, so you wouldn't need an interface that all the elements in the list would have to extend, rather have them wrapped in some object, and unwrapped and casted later.

When you specify here <T> I think the whole idea of having a DisplayableItem loses it's intention and is meaningless, could be removed.

tomaszpolanski: The purpose of this PR is not to remove `DisplayableItem` yet. It was to fix it a bit as omitting the generic part was making it type unsafe.
In the future, we could think about how this could be done better in statically typed way

luciapayo: IMO when you want to have different items in one adapter you have two approaches:
- Make it type unsafe with the wrapping approach here: the good about this is that you remove the boiler plate of having to create interfaces.
- Create interfaces for all the elements in a given list

Not sure what option is best, but maybe we should discuss about it before deciding

tomaszpolanski: `RecyclerViewAdapter<T>` should be generic. The difference should be that it should store union type and not only `Sound` item.
This PR is first step to make it happen (making it now a bit type safe)

luciapayo: Ah, I didn't know you could do that on kotlin. Sounds promising. I'll finish the review later

app/src/main/java/com/futurice/freesound/feature/search/SearchFragmentModule.java
```diff
@@ -52,11 +53,12 @@ static SearchFragmentViewModel provideSearchFragmentViewModel(SearchDataModel se
 
     @Provides
     @FragmentScope
-    RecyclerViewAdapter provideRecyclerAdapter(ItemComparator itemComparator,
-                                               Map<Integer, ViewHolderFactory> factoryMap,
-                                               Map<Integer, ViewHolderBinder> binderMap,
-                                               SchedulerProvider schedulerProvider) {
-        return new RecyclerViewAdapter(itemComparator, factoryMap, binderMap, schedulerProvider);
+    RecyclerViewAdapter<Sound> provideRecyclerAdapter(ItemComparator itemComparator,
+                                                      Map<Integer, ViewHolderFactory> factoryMap,
```
luciapayo: Shouldn't this be `ViewHolderFactory<Sound>`?

tomaszpolanski: Nope, neither `ViewHolderFactory` nor the thing it creates is a generic

PR #138, opened 2017-07-03
MVI HomeFragment

peter-tackage:
I'm looking for some high-level feedback on the general mechanism used in `HomeFragmentViewModel2` and how that is consumed by the `HomeFragment`.

# The General Idea:

- The general idea is to remove UI inconsistency and hard-to-reason-with state that is a characteristic of the existing MVVM pattern in use. This includes scenario where multiple `Observable`s emit due to a change, making it difficult to predict the resulting contents of the `View`.

- The ViewModel uses Google's new `ViewModel` component which survives configuration change 
    - Its scope exceeds that of its associated View.
    - All input from the data layer to the `ViewModel` is channeled through one `LiveData` object exposed by `uiModel()`

- The View binds to the ViewModel using its `UiBinder`.
    - This indirectly holds a subscription which:
        - Sends the UiModel from the ViewModel to the View's renderer
        - Send the `UiEvent`s (clicks etc) from the View to the ViewModel's `uiEvents` function 
    - This subscription is created/destroyed on View start/stop (**using Lifecycle aware components**)
    - The `UiBinder` has the same scope as its host View.

- Inside the ViewModel: 
  - On instantiation the ViewModel subscribes to an `Observable` which transforms `UiEvent`s into `UiModel`s and updates the `LiveData`.
    - The View `UiEvent`s are mapped to a `Action` event.
    - The `Action` is transformed into a `Result` event.
    - The `Result` is then given to the `reduce` function which determines how to update the *overall view state based upon the change and the previous state*.
  - The `uiModel` `LiveData` exposes this to view layer consumers:       
   - The `UiBinder` consumes these and displays the state
  - The `ViewModel` performs unsubscribe from its internal `Disposable`

# Component Roles

- ViewModel
- Interactor
- Repository
  - `get`
  - `getStream`
  - `refresh`
- CacheStore

# Things I'm trying to improve:

## Render logic could be moved outside of View

It might be a bit more MVPresenter, but the Renderer logic for each View could be moved outside of the View and used to test in a more understandable way - e.g. when a UserUiModel object is non-null, then then `showUser` is called, else `hideUser` is called. This Renderer logic would have to remain View scoped, rather than ViewModel scoped.

## Don't allow unexpected errors to propagate to the UiBinder (and crash the app)

The `UiBinder` uses LiveData, which cannot propagate errors when fed by an `Observable`. Any such attempt will cause the app to crash. To avoid this, we could represent unexpected errors as a state. This seems a little heavy handed though.

## ~~What's the point of using the `ViewModel` Architecture Component??~~

~~I guess, currently the pattern doesn't need it. Although I'm looking at revisiting this and possibly changing the pattern because of the next point...~~

## ~~Long running operations unsubscribe by the ViewModel~~

~~There's no caching or stored state in the ViewModel, so on configuration change any in-flight requests being handled by the ViewModel are unsubscribed and lost.~~

~~Not every ViewModel will handle these types of requests, so the current solution would be for the ViewModel to dispatch the request to a component which will run "hot" (without subscribers). Arguably, this is a fairly reasonable solution already.~~


app/src/main/java/com/futurice/freesound/feature/home/HomeFragmentDataEvents.kt
```diff
@@ -0,0 +1,31 @@
...
+import java.util.concurrent.TimeUnit
+
+// This could probably just be a file, it would nice to have symmetric way of doing this
+internal class HomeFragmentDataEvents(private val userDataModel: UserDataModel) {
+
+    fun dataEvents(): Observable<Fragment.DataEvent> {
+        return userDataModel.homeUser
+                .delay(5, TimeUnit.SECONDS) // for testing purposes!!!!
+                .map { Fragment.DataEvent.UserDataEvent(it) as Fragment.DataEvent }
+                .toObservable()
```
tomaszpolanski: Why do you change it to Obervable?

app/src/main/java/com/futurice/freesound/feature/home/HomeFragmentViewModel2Factory.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package com.futurice.freesound.feature.home;
+
+import com.futurice.freesound.feature.common.scheduling.SchedulerProvider;
+
+import android.arch.lifecycle.ViewModel;
+import android.arch.lifecycle.ViewModelProvider;
+
+class HomeFragmentViewModel2Factory implements ViewModelProvider.Factory {
```
tomaszpolanski: This one might be much better with Kotlin and reified

app/src/main/java/com/futurice/freesound/mvi/BaseViewModel.kt
```diff
@@ -0,0 +1,75 @@
...
+package com.futurice.freesound.mvi
+
+import com.futurice.freesound.feature.common.scheduling.SchedulerProvider
+import com.jakewharton.rx.replayingShare
+import io.reactivex.Observable
+import io.reactivex.disposables.SerialDisposable
+import io.reactivex.processors.PublishProcessor
+import timber.log.Timber
+
+abstract class BaseViewModel<E, D, M, C>(
```
tomaszpolanski: Maybe we should use more descriptive names

app/src/main/java/com/futurice/freesound/feature/home/HomeFragmentViewModel2Factory.java
peter-tackage: I found a good way of doing that, but I might have to juggle it around at bit because you can't call reified functions from Java where this factory is called from.

app/src/main/java/com/futurice/freesound/feature/home/HomeFragmentDataEvents.kt
peter-tackage: The general pattern would be for the VM to take an `Observable<DataEvents>`, in this case it's only a `Single`, but I wanted to try and emulate the more general case.

app/src/main/java/com/futurice/freesound/mvi/BaseViewModel.kt
peter-tackage: Most likely, yes. I'm currently thinking that it' over parameterized. Perhaps only `E`, `D` and `M` are necessarily. If nothing else, they are the more important of the interactions, so perhaps it's possible to hide `C`.

peter-tackage:
@tomaszpolanski There's been a bit of rework, if you've got some time to take look.

PR #139, opened 2017-07-03
Suggestion on how we can do arrange builder

tomaszpolanski:


peter-tackage:
Looks similar to what I was thinking: https://gist.github.com/peter-tackage/7ce277b6c77a1a68411207c909d672b4

tomaszpolanski:
I would avoid too complex inner class structures, we can try it out case by case basis. For the class that I am testing, I would say that's enough.
For more complex scenarios we can have something more sophisticated.

peter-tackage:
Yeah, I played with `Act` and `Assert` as a part of the DSL but there's too many combinations of structures to find a common nesting pattern. We're better off, like you said of just starting with `Arrange` and seeing what comes of it.

app/src/test/java/com/futurice/freesound/network/api/DefaultFreeSoundApiServiceTest.kt
```diff
@@ -0,0 +1,198 @@
...
+        defaultFreeSoundApiService.search("query")
+                .test()
+                .assertError(ERROR)
+    }
+
+    fun arrange(init: Arrangement.() -> Unit) = Arrangement().apply { init() }
+
+    inner class Arrangement {
+
+        fun search(init: () -> SoundSearchResult): Any =
```
tomaszpolanski: I wanted to avoid having a `{ return... }` function and preferred `=` function.
But `when` returns something that is not a `Unit`.
I am not yet sold on my solution, so if you do not like it, I can change it

app/src/test/java/com/futurice/freesound/network/api/DefaultFreeSoundApiServiceTest.kt
```diff
@@ -0,0 +1,198 @@
...
+import org.junit.Test
+import org.mockito.ArgumentMatchers.*
+import org.mockito.Mock
+import org.mockito.Mockito.`when`
+import org.mockito.Mockito.verify
+import org.mockito.MockitoAnnotations
+
+class DefaultFreeSoundApiServiceTest {
+
+    private val TEST_CLIENT_ID = "clientId"
```
tomaszpolanski: In tests I do not see need for constants to be in companion objects

peter-tackage: I was thinking the same, because the values are not accessed outside the class. I will assume for now that there's no significant performance difference.

app/src/test/java/com/futurice/freesound/network/api/DefaultFreeSoundApiServiceTest.kt
```diff
@@ -0,0 +1,193 @@
...
+                .test()
+                .assertError(ERROR)
+    }
+
+    fun arrange(init: Arrangement.() -> Unit) = Arrangement().apply(init)
+
+    inner class Arrangement {
+
+        fun search(init: () -> SoundSearchResult): Any =
+                `when`(freeSoundApi.search(anyString(), any<String>(), any(SoundFields::class.java)))
```
peter-tackage: Could we make this `any(SoundFields::class.java)` become:

```kotlin
    inline fun <reified T> any() : T = Mockito.any(T::class.java)
```

so call site is: ` any<SoundFields>()`

There's also this: https://github.com/nhaarman/mockito-kotlin/ which I has these already (I used this in kotlin-options).

app/src/test/java/com/futurice/freesound/network/api/DefaultFreeSoundApiServiceTest.kt
peter-tackage: It's probably terrible, but:

```kotlin
    fun Any?.toUnit() : Unit {}
```
and call site:
```kotlin
        fun search(init: () -> SoundSearchResult)  =
            `when`(freeSoundApi.search(anyString(), any<String>(), any<SoundFields>()))
                    .thenReturn(Single.just(init())).toUnit()
```

I wouldn't say it's better than just surrounding with`{}`.

peter-tackage: Alternatively, if the formatter didn't wrap on the braces, you would probably get the same visual representation:

```kotlin
fun search(init: () -> SoundSearchResult)  
            { `when`(freeSoundApi.search(anyString(), any<String>(), any<SoundFields>()))
                    .thenReturn(Single.just(init())) }
```

peter-tackage: @tomaszpolanski I think the best solution result in not exposing `Any` from the Arrange DSL API calls. I would say, either surround in braces or use the `toUnit` (`asUnit` possibly?).

PR #141, opened 2017-07-05
Migrating partialy home to kotlin

tomaszpolanski:


peter-tackage:
Out of curiosity, how much of this is manual vs automated conversion?

app/src/main/java/com/futurice/freesound/feature/home/HomeActivity.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    @Inject
+    internal lateinit var homeViewModel: HomeViewModel
+
+    private val dataBinder = SimpleDataBinder()
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_home)
+        if (savedInstanceState == null) {
```
peter-tackage: Use safe `let`? 

It's probably also a style question for these simple cases.

app/src/main/java/com/futurice/freesound/feature/home/HomeActivity.kt
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        return when (item.itemId) {
+            R.id.action_about  -> true
+            R.id.action_search -> {
+                homeViewModel.openSearch()
+                true
+            }
+            else               -> super.onOptionsItemSelected(item)
```
peter-tackage: Formatting. (Also run the auto-formatter, there's some other minor formatting issues).

app/src/main/java/com/futurice/freesound/feature/home/HomeFragmentViewModel.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.futurice.freesound.network.api.model.User
+import com.futurice.freesound.viewmodel.SimpleViewModel
+import io.reactivex.Single
+
+internal class HomeFragmentViewModel(userDataModel: UserDataModel) : SimpleViewModel() {
+
+    private val homeUser: Single<User> = userDataModel.homeUser.cache()
+
+    val image: Single<String>
+        get() = homeUser.map { it.avatar().large() }
```
peter-tackage: Can you explain the benefits of using property notation here, rather than just having methods named `image`, `username` etc.

app/src/main/java/com/futurice/freesound/feature/home/HomeActivity.kt
tomaszpolanski: let does something when the object is not null, here we have the reverse

app/src/main/java/com/futurice/freesound/feature/home/HomeActivity.kt
tomaszpolanski: Actually the formatting is good, it aligns with other cases

app/src/main/java/com/futurice/freesound/feature/home/HomeFragmentViewModel.kt
tomaszpolanski: A property is a function that does not accept any arguments and does not change the internal state of the object, here is exactly the case

app/src/main/java/com/futurice/freesound/feature/home/HomeActivity.kt
peter-tackage: Manual alignment like this adds maintenance effort each time the `when` changes. I can see potential benefit for long table-like declarations in public APIs, but not here.

app/src/main/java/com/futurice/freesound/feature/home/HomeActivity.kt
peter-tackage: Of course. I'm so used to seeing null checks as a potential code smell. Is there an equivalent for safe `let` when you want to perform an action if null?

peter-tackage: I guess ` savedInstanceState ?: addHomeFragment()` would work. It's more concise, but I'm not convinced about readability.

tomaszpolanski: Exactly, when I was converting it I made a mistake of thinking that we need not null. Probably we should keep it explicit if there is not a common way of doing it

app/src/main/java/com/futurice/freesound/feature/home/HomeActivity.kt
tomaszpolanski: The thing is that it was not a manual alignment, autoformatting did that

app/src/main/java/com/futurice/freesound/feature/home/HomeActivity.kt
peter-tackage: A single line if `(savedInstanceState == null) addHomeFragment()` would be preferable, I think.

app/src/main/java/com/futurice/freesound/feature/home/HomeActivity.kt
peter-tackage: I guess we're using different formatters. I'm not really sold on this style, it seems to go against my view that the general trend with Kotlin is writing compact code and reducing superfluous characters. I wouldn't do it in Java either though. I usally read the `when` block vertically (to find the relevant case), then horizontally (to find the associated action).

app/src/main/java/com/futurice/freesound/feature/home/HomeFragmentViewModel.kt
peter-tackage: Ok, let's try this property approach out and see how it goes. My only worry is that there's a bit more required knowledge (regarding properties in Kotlin) for people who are more used to Java.

Issue #144, opened 2017-07-11
Add tests for Json Parsing


peter-tackage:
We currently have no tests for our API model JSON parsing, meaning that changes to mapping libraries etc cannot be verified without running the app.

We should have tests for the basic parsing of the model objects.

peter-tackage:
This is generally the type of testing that I had in mind: https://github.com/peter-tackage/soundmap/tree/master/app/src/test

Start with the happy path for a single model type in the first PR and go from there.

Kotlin has some nice helpers for reading in files if that's the approach you want to use.

joan-domingo:
👍 will follow that one

PR #147, opened 2017-07-29
Migrating model step 2

tomaszpolanski:
In this PR I've simplified the way how we use Parcelable with data classes, migrated Sound model (without changing nullability) and improved moshi conversion


peter-tackage:
@joan-domingo - this change removes Gson, so it could be relevant for #144

app/src/main/java/com/futurice/freesound/network/api/model/mapping/DateAdapter.kt
```diff
@@ -0,0 +1,26 @@
...
+
+package com.futurice.freesound.network.api.model.mapping
+
+import com.squareup.moshi.FromJson
+import java.util.*
+
+class DateAdapter {
+
+    @FromJson fun fromJson(json: String): Date {
+        val parsableDate = if (json.endsWith("Z")) json else json.plus("Z")
```
peter-tackage: Why can't we use `Rfc3339DateJsonAdapter.java`?

app/src/main/java/com/futurice/freesound/network/api/ApiNetworkModule.java
```diff
@@ -69,20 +65,12 @@ static FreeSoundApi provideFreeSoundApi(@Named(API_URL_CONFIG) String url,
...
-    }
-
-    @Provides
-    @Singleton
-    @ForFreeSoundApi
     static Moshi provideMoshi() {
         return new Moshi.Builder()
                 .add(new KotlinJsonAdapterFactory())
                 .add(new GeoLocationJsonAdapter())
+                .add(new DateAdapter())
```
peter-tackage: I'm confused as to why we need both these

app/src/main/java/com/futurice/freesound/network/api/model/mapping/DateAdapter.kt
tomaszpolanski: Because the date format that freesound provides is not according to the standard. The `Z` from the end is missing

app/src/main/java/com/futurice/freesound/network/api/ApiNetworkModule.java
tomaszpolanski: I've spent like 5h on the issue. If one adapter is not included, we get json parsing error. If other is not included, it fails silently.

peter-tackage: Any idea how it was working correctly before with Gson? I don't remember having a special date deserializaer.

peter-tackage: It's funny how the code from the `Rfc3339DateJsonAdapter` is originally from Jackson, then Gson, then Moshi.... and now we have our own copy.

tomaszpolanski: In Gson we were giving time format as a string, Moshi does not have this feature

peter-tackage: When using only `Rfc3339DataJsonAdapter`, the parser correctly throws on this code:

```java
   // extract timezone
      if (date.length() <= offset) {
        throw new IllegalArgumentException("No time zone indicator");
      }
```

and then throws from here:

```java
      throw new JsonDataException("Not an RFC 3339 date: " + date);
```

Weird that it's failing silently afterwards: no stacktrace, no log, no error indicator. The error is definitely propagated out of Retrofit though.

peter-tackage: Ok, I do see the search error pop-up, but we should at least log non-network based exceptions. Although that should be done at the Retrofit/OkHttp level, not here (and in a different PR).

peter-tackage: Could we do something like:

```kotlin
class OurDateAdapter(private val delegate: Rfc3339DateJsonAdapter)  {

    @FromJson fun fromJson(json: String): Date {
        val parsableDate = if (json.endsWith("Z")) json else json.plus("Z")
        return delegate.fromJson(parsableDate)!! // TODO do the null safety better
    }

}
```

peter-tackage: Ok this works:

```kotlin
    @Provides
    @Singleton
    @ForFreeSoundApi
    static Moshi provideMoshi() {
        return new Moshi.Builder()
                .add(new KotlinJsonAdapterFactory())
                .add(new GeoLocationJsonAdapter())
                .add(new OurDateAdapter(new Rfc3339DateJsonAdapter()))
                .build();
    }
```
and 

```kotlin
class OurDateAdapter(private val delegate: Rfc3339DateJsonAdapter) {

    @FromJson fun fromJson(json: String): Date {
        val parsableDate = if (json.endsWith("Z")) json else json.plus("Z")
        return delegate.fromJson("\"" + parsableDate + "\"")
                .takeIf { it != null }
                ?: throw JsonDataException("Unable to deserialize Date from: $json")
    }

    @ToJson fun toJson(date: Date): String {
        return delegate.toJson(date)
    }

}
```

peter-tackage: I can push that if you're happy with it.

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
```diff
@@ -0,0 +1,142 @@
...
+
+import android.os.Parcel
+import com.futurice.freesound.common.utils.*
+import com.squareup.moshi.Json
+import java.util.*
+
+/**
+ * Refer to: http://www.freesound.org/docs/api/resources_apiv2.html#sound-resources
+ */
+data class Sound(
```
peter-tackage: I would like to make the fields here non-nullable, perhaps with a sensible default, but the Freesound API is a little annoying in that you have to request specific fields you want in the response object, which would be great if we had dynamic typing, but not so good for us here. In any case, it's out of scope for this PR.

app/src/main/java/com/futurice/freesound/common/utils/Parcels.kt
```diff
@@ -0,0 +1,92 @@
...
+package com.futurice.freesound.common.utils
+
+import android.os.Parcel
+import android.os.Parcelable
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.util.*
+import kotlin.collections.ArrayList
+
+interface KParcelable : Parcelable {
```
peter-tackage: Could you please add a sprinkling of comments here (it doesn't have to be full Javadoc), so it's clearer what's going on.

app/src/main/java/com/futurice/freesound/network/api/model/GeoLocation.kt
```diff
@@ -17,28 +17,22 @@
...
 
-import android.os.Parcelable
-
-// Using https://github.com/nekocode/android-parcelable-intellij-plugin-kotlin for Parcelable
 data class GeoLocation(
-        val latitude: Double,
-        val longitude: Double) : Parcelable {
+    val latitude: Double,
+    val longitude: Double) : KParcelable {
     constructor(parcel: Parcel) : this(
```
peter-tackage: `GeoLocation` ctor can be `private`

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
```diff
@@ -0,0 +1,142 @@
...
+        // The username of the uploader of the sound.
+        val username: String?,
+        // Thumbnail image URLs of the waveform/spectral plot
+        val images: Image,
+        // Preview sounds URLs
+        val previews: Preview?,
+        // Duration in seconds
+        val duration: Float?,
+        val created: Date) : KParcelable {
+    constructor(parcel: Parcel) : this(
```
peter-tackage: `Sound` ctor can be `private`

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        @Json(name = "waveform_l")
+        val largeSizeWaveformUrl: String?,
+
+        @Json(name = "spectral_m")
+        val medSizeSpectralUrl: String?,
+
+        @Json(name = "spectral_l")
+        val largeSizeSpectralUrl: String?) : KParcelable {
+    constructor(parcel: Parcel) : this(
```
peter-tackage: `Image` ctor can be `private`

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        @Json(name = "preview-hq-mp3")
+        val highQualityMp3Url: String?,
+
+        @Json(name = "preview-lq-ogg")
+        val lowQualityOggUrl: String?,
+
+        @Json(name = "preview-hq-ogg")
+        val highQualityOggUrl: String?) : KParcelable {
+    constructor(parcel: Parcel) : this(
```
peter-tackage: `Preview` ctor can be `private`

app/src/main/java/com/futurice/freesound/common/utils/Parcels.kt
```diff
@@ -0,0 +1,92 @@
...
+// Creator factory functions
+
+inline fun <reified T> parcelableCreator(
+    crossinline create: (Parcel) -> T) =
+    object : Parcelable.Creator<T> {
+        override fun createFromParcel(source: Parcel) = create(source)
+        override fun newArray(size: Int) = arrayOfNulls<T>(size)
+    }
+
+inline fun <reified T> parcelableClassLoaderCreator(
```
peter-tackage: This doesn't appear to be used; is it needed?

app/src/main/java/com/futurice/freesound/common/utils/Parcels.kt
peter-tackage: Also tests, or at least tests for the Parcel generation from the data model perspective which would exercise these.

peter-tackage: Also attribution.

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    companion object {
+        @JvmField val CREATOR = parcelableCreator(::Sound)
+    }
+}
+
+data class Image(
+
+        @Json(name = "waveform_m")
+        val medSizeWaveformUrl: String,
```
peter-tackage: Why is this non-null and the others nullable?

tomaszpolanski: Because in one of the vm you assume that this is not null. The app would be crashing if that was null

app/src/main/java/com/futurice/freesound/common/utils/Parcels.kt
tomaszpolanski: For that, we would need espresso, but I cannot see it in the project (?!). If that is the case, after this PR I will add espresso

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
peter-tackage: We'll have to define a more consistent approach (another PR).

PR #148, opened 2017-08-09
Use joinToString extension rather than own implementation

peter-tackage:
Kotlin has an extension method to achieve the same thing. I don't want to convert the `SoundFields` class to Kotlin just yet, so I'm only updating the implementation.


app/src/main/java/com/futurice/freesound/network/api/model/RequestHelper.java
```diff
@@ -20,7 +20,7 @@
 
 import android.support.annotation.NonNull;
 
-import java.util.Iterator;
+import kotlin.collections.CollectionsKt;
```
tomaszpolanski: Why didn't you just convert the file to kotlin and use `joinToString` as an extension method?

peter-tackage: No good reason, this PR just started as a demo of calling the Kotlin stdlib from Java.

PR #151, opened 2017-08-13
Migrating search activity to kotlin

tomaszpolanski:


app/src/main/java/com/futurice/freesound/common/utils/StandardExtensions.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and limitations under the License.
+ *
+ */
+
+package com.futurice.freesound.common.utils
+
+inline fun <T> T?.ifNull(block: () -> Unit): Unit = if (this == null) block() else Unit
```
peter-tackage: It's not one line, but how about:

```kotlin
inline fun <T> T?.ifNull(block: () -> Unit) {
    if (this == null) block()
}
```

tomaszpolanski: Might be a bit cleaner

PR #152, opened 2017-08-13
Remove Butterknife reference from SearchActivity

peter-tackage:
There was still a Butterknife reference in the Activity.


peter-tackage:
Also @tomaszpolanski  can you check your formatting settings as we seem to have some differences. screen shot 2017-08-13 at 21 33 11

tomaszpolanski:
Could you export the style and add it to the repo?

PR #153, opened 2017-08-18
Migrating search fragment to kotlin

tomaszpolanski:


app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.kt
```diff
@@ -0,0 +1,128 @@
...
+    override fun createComponent(): SearchFragmentComponent =
+            (activity as SearchActivity).component()
+                    .plusSearchFragmentComponent(BaseFragmentModule(this))
+
+    override fun viewModel(): ViewModel = searchFragmentViewModel
+
+    override fun dataBinder(): DataBinder = dataBinder
+
+    private fun handleResults(sounds: List<DisplayableItem<Sound>>?) {
+        if (sounds != null) showResults(sounds) else showNothing()
```
peter-tackage: Didn't you create an extension to handle this situation?

app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.kt
```diff
@@ -0,0 +1,128 @@
...
+    @Inject
+    internal lateinit var schedulerProvider: SchedulerProvider
+
+    private val dataBinder = object : DataBinder {
+
+        override fun bind(disposables: CompositeDisposable) {
+            disposables += searchFragmentViewModel.soundsOnceAndStream
+                    .subscribeOn(schedulerProvider.computation())
+                    .observeOn(schedulerProvider.ui())
+                    .subscribe({ handleResults(it.orDefault { null }) })
```
peter-tackage: Do you have a bigger picture in mind which involves not exposing `Option` to too many parts of the code? Why not just push down the `Option` to the `handleResults` method as was done previously?

tomaszpolanski: I think we should slowly remove my options and replace those with yours. Here, the handleResults logic is a bit simpler when using the null instead of option

app/src/main/java/com/futurice/freesound/feature/search/SearchFragment.kt
tomaszpolanski: I only created `ifNull` extension. Not yet sure if we need one for null and one for not as we can do it with pattern matching

PR #155, opened 2017-09-10
Add RxLint

peter-tackage:


app/src/main/java/com/futurice/freesound/feature/search/SearchFragmentViewModel.java
```diff
@@ -56,6 +57,7 @@
         return searchDataModel.getSearchStateOnceAndStream()
                               .map(SearchState::results)
                               .map(it -> it.map(SearchFragmentViewModel::wrapInDisplayableItem))
+                              .delay(1, TimeUnit.SECONDS)
```
peter-tackage: This should trigger the lint warning because I haven't used `delay` with a Scheduler.

peter-tackage:
@maritoast I'll wait for the non-snapshot version of the library to be released before merging this.

PR #162, opened 2017-11-19
Use Jonty

peter-tackage:
I've replaced the handcrafted list of fields with a Kotlin based annotation processor which creates the appropriate list in a Kotlin file. Review of the annotation processor: https://github.com/peter-tackage/jonty also welcomed/encouraged.

I should add that generating Kotlin via an annotation processor for Android brings out the worst in the build system; hence the many long explanations littered over the `build.gradle`.


app/build.gradle
```diff
@@ -40,6 +41,22 @@ android {
...
+        //
+        // Note: This option has also been targeted for deprecation by the Android Gradle plugin.
+        //
+        // Refer: https://github.com/google/auto/issues/268
+        javaCompileOptions {
+            annotationProcessorOptions {
+                includeCompileClasspath false
+            }
+        }
+
```
maritoast: line

app/src/main/java/com/futurice/freesound/network/api/DefaultFreeSoundApiService.java
```diff
@@ -68,7 +69,10 @@
     @Override
     @NonNull
     public Single<SoundSearchResult> search(@NonNull final String query) {
-        return freeSoundApi.search(get(query), null, SoundFields.BASE);
+        Sound_JontyFielder.INSTANCE.getFields();
```
maritoast: I think you removed this already

peter-tackage:
@maritoast I'd still like to fix the Kotlin companion object field bug, which will require changes to Jonty.

PR #163, opened 2017-12-06
Making fields some fields coming from the backend not null

tomaszpolanski:


app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.futurice.freesound.network.api.model.Sound
+import com.futurice.freesound.viewmodel.SimpleViewModel
+import com.google.auto.factory.AutoFactory
+import com.google.auto.factory.Provided
+import io.reactivex.Observable
+import io.reactivex.Single
+import polanski.option.Option
+import java.text.DateFormat
+
+@AutoFactory
```
peter-tackage: Hmm, I thought AutoFactory didn't work with Kotlin.

tomaszpolanski: It seems that it work now :)

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
```diff
@@ -0,0 +1,98 @@
...
+internal class SoundItemViewModel(private val sound: Sound,
+                                  private @Provided val navigator: Navigator,
+                                  private @Provided val audioPlayer: AudioPlayer,
+                                  private @Provided val freeSoundApiService: FreeSoundApiService) : SimpleViewModel() {
+
+    private val thumbnail: String
+        get() = sound.images.medSizeWaveformUrl
```
peter-tackage: What the benefit of using `get()` syntax here, as opposed to using just `private val thumbnail: String = sound.images.medSizeWaveformUrl`?

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    private val currentPercentage: Observable<Option<Int>>
```
peter-tackage: Not sure about the best practice, but it's also possible to infer the `currentPercentage` type.

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
```diff
@@ -0,0 +1,98 @@
...
+                .map { positionMs -> toPercentage(positionMs, sound.duration) }
+                .map { Option.ofObj(it) }
+
+    fun thumbnailImageUrl(): Single<String> = Single.just(thumbnail)
+
+    fun name(): Single<String> = Single.just(sound.name)
+
+    fun userAvatar(): Single<String> =
+            freeSoundApiService.getUser(sound.username)
+                    .map { (_, _, avatar) -> avatar.medium }
```
peter-tackage: Is decomposing the user type here actually any more readable than just mapping via:
`map { user -> user.avatar.medium }`

peter-tackage: If we have to add more properties to the User class, then `.map { (_, _, avatar) -> avatar.medium }` will have to change, which to me, is unnecessary.

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
```diff
@@ -0,0 +1,98 @@
...
+                    .map { it.time }
+                    .map { d -> DateFormat.getDateInstance().format(d) }
+
+    fun username(): Single<String> = Single.just(sound.username)
+
+    fun description(): Single<String> = Single.just(sound.description)
+
+    fun duration(): Single<Int> =
+            Single.just(sound.duration)
+                    .map { duration -> Math.ceil(duration.toDouble()).toInt() }
```
peter-tackage: There's also a new `kotlin.math` package in 1.2

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
```diff
@@ -0,0 +1,98 @@
...
+        navigator.openSoundDetails(sound)
+    }
+
+    fun toggleSoundPlayback() {
+        audioPlayer.togglePlayback(
+                PlaybackSource(from(sound.id),
+                        sound.previews.lowQualityMp3Url))
+    }
+
+    private fun progressOrNothing(playerState: PlayerState): Observable<Option<Int>> {
```
peter-tackage: Thoughts on making these helper functions be extension methods instead?

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
```diff
@@ -28,51 +28,51 @@ import java.util.*
 @Fieldable
 data class Sound(
```
peter-tackage: The tricky thing we have here is that depending which field are actually requested from the backend, then the properties can actually be nullable from the API. Strictly speaking, there are no inherent non-nullable properties for `Sound`. In our current usage, we are using all the fields generated from `@Fieldable`, but there's no reason why another part of the API service layer could not make a request with a subset of those fields, in which case we would expect the field to be null.

peter-tackage: Perhaps we should allow the fields to be nullable in the API response mapped class (currently `Sound`) but then have another data type which is used by the remainder of the app.  We should rename `Sound` to be `SoundResponse` and then redefine a `Sound` class with non-nullable field for use by clients of the API service. Add a mapping layer which enforces the intended nullability.

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
```diff
@@ -28,51 +28,51 @@ import java.util.*
...
         // Thumbnail image URLs of the waveform/spectral plot
         val images: Image,
         // Preview sounds URLs
-        val previews: Preview?,
+        val previews: Preview,
         // Duration in seconds
-        val duration: Float?,
+        val duration: Float,
         val created: Date) : KParcelable {
     private constructor(parcel: Parcel) : this(
```
peter-tackage: Was there some reason why we are not using the Android extensions @Parcelize? Should available since 1.1.4 with:

```groovy
androidExtensions {
    experimental = true
}
```

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
peter-tackage: I think that would still fit within the scope of this PR.

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
tomaszpolanski: In this case, I would leave it for the readability sake

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
tomaszpolanski: That one is mainly for usage for JS. I think it can stay for now like this

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
tomaszpolanski: It was not stable enough in 1.1.4. But since 1.2 we can try it out

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
tomaszpolanski: In the documentation, there is nothing mentioned about nullability.
That is why I've tried out the app with the current nullability setup and worked properly.
I have added a Timber logging to where we are passing them and we should have an eye on that via crashalitycs.
I would not like to add nullability just because we do not know

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
tomaszpolanski: But that would be in different PR

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
tomaszpolanski: Could you give an example how it would make it better?

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
peter-tackage: When requesting a sound from the API you need to request the fields that you wish to be populated (Jonty and all that) - this means that in the JSON response any fields not requested are missing and hence would be null when mapped to `Sound.kt`.

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
peter-tackage: Yep different PR. I've raised one - I like to live on the edge.

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
tomaszpolanski: Ok, let's discuss this week the Fieldable and your idea

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
peter-tackage: Only that a private extension method might read better from the call site, than when the receiver is a parameter:

So they would look a bit like:
`playerState.progressOrNothing()` 
`playerState.source.isThisSound()`
`position.ToPercentage()`

peter-tackage: For private extensions there's not much of a downside in terms of perceived instance method pollution.

app/src/main/java/com/futurice/freesound/feature/search/SoundItemViewModel.kt
peter-tackage: Yeah, unless the extension methods are particularly useful, the Java SDK is no doubt sufficient.

app/src/main/java/com/futurice/freesound/network/api/model/Sound.kt
peter-tackage: @tomaszpolanski Let's progress with what we have and hopefully we have the problem of devising a new solution later.

PR #166, opened 2017-12-09
Use @Parcelize

peter-tackage:
It's still experimental, but worth exploring because it will save 100s of LOC. Some documentation here: https://github.com/Kotlin/KEEP/blob/master/proposals/extensions/android-parcelable.md


peter-tackage:
@tomaszpolanski I've added a test for `Sound.kt`. Because there's no  `CREATOR` object visible in the code, I've had to let Android perform the serialization/deserialization via Bundle.

joan-domingo:
Does it work fine? I haven't tried with Kotlin 1.2, but in previous versions there was an issue parceling Date: https://stackoverflow.com/questions/46592138/nosuchmethoderror-using-parcelize-annotation-in-kotlin

peter-tackage:
@joan-domingo I have a test which uses `Date` and passes. Also the [Date issue](https://youtrack.jetbrains.com/issue/KT-19749) in questions seems to be fixed. It's still experimental, so who knows, perhaps they could change the API or remove it. But given the simplicity of use - it's a great improvement over the old boilerplate.

peter-tackage:
@joan-domingo, @tomaszpolanski has pointed out that the test I wrote does effectively test the marshalling/unmarshalling - in fact `sound` and `result` are the same instance, hence will always pass the equality assertion.

PR #169, opened 2017-12-09
Converting images package to kotlin

tomaszpolanski:


app/src/main/java/com/futurice/freesound/feature/home/HomeFragment.kt
```diff
@@ -53,7 +53,7 @@ class HomeFragment : BindingBaseFragment<HomeFragmentComponent>() {
                 .observeOn(schedulerProvider.ui())
                 .subscribe({
                                picasso.load(it)
-                                   .transform(PicassoTransformations.circular())
+                                   .transform(circular())
```
peter-tackage: I'm not sure that this namespace change is for the better.

app/src/main/java/com/futurice/freesound/feature/images/PicassoTransformations.kt
```diff
@@ -13,24 +13,14 @@
...
 
-package com.futurice.freesound.feature.images;
+package com.futurice.freesound.feature.images
 
-import com.futurice.freesound.common.InstantiationForbiddenError;
-import com.squareup.picasso.Transformation;
+import com.squareup.picasso.Transformation
 
-import android.support.annotation.NonNull;
+private val CIRCULAR = RoundEdgeTransformation()
```
peter-tackage: Would there be a benefit to making this an object?

peter-tackage: (perhaps in RoundEdgeTransformation.kt)

app/src/main/java/com/futurice/freesound/feature/home/HomeFragment.kt
tomaszpolanski: The change is because `circular` is in "no" class. And I dont think that using it like `com.futurice.freesound.feature.images.circular())` here would benefit it

peter-tackage: That being the case, I'll rename `circular` to `circularTransformation` to make it clearer.

peter-tackage:
@tomaszpolanski Merge once we've/they've sorted out the Travis build request issues.

